PR #33, opened 2021-09-17
create dir to store the creds

didrocks:


didrocks:
Superseeded now that main is rebased.

PR #41, opened 2021-10-13
Fix update db

didrocks:
Do not rely on inputs parameters for updating db and printing notice as daily build don’t have any.


jibel (approved):
lgtm.

PR #43, opened 2021-10-14
Turned 'common' folder into a Module

cnihelton:
While working with the Distro launcher code on Windows and Visual Studio I stumbled upon a few issues to prepare the build system. Those issues were due Windows not able to understand the symlinks from the 'wsl-builder/common' folder which exists inside 'prepare-assets' and 'prepare-build'. Thus, 'prepare-build' wouldn't compile on Windows.

With this PR I aim to convert the 'common' folder into a module which can be imported by 'prepare-assets' and 'prepare-build' and enable users working on Windows to prepare the build system without the need for spawning a VM or WSL just for that.

That required a few case changes here and there to make the function and type names visible by the importer modules.

### DIFF
--- a/wsl-builder/prepare-assets/assets.go
+++ b/wsl-builder/prepare-assets/assets.go
@@ -14,8 +14,8 @@ import (
 	"text/template"
 
 	shutil "github.com/termie/go-shutil"
+	"github.com/ubuntu/WSL/wsl-builder/common"
 	"gopkg.in/gographics/imagick.v2/imagick"
-	"github.com/ubuntu/WSL-DsitroLauncher/wsl-builder/common"
 )
 
 // updateAssets orchestrates the distro launcher metadata and assets generation from a csv file path.
--- a/wsl-builder/prepare-assets/go.sum
+++ b/wsl-builder/prepare-assets/go.sum
@@ -99,10 +99,8 @@ github.com/kisielk/errcheck v1.1.0/go.mod h1:EZBBE59ingxPouuu3KfxchcWSUPOHkagtvW
 github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+oQHNcck=
 github.com/konsorten/go-windows-terminal-sequences v1.0.1/go.mod h1:T0+1ngSBFLxvqU3pZ+m/2kptfBszLMUkC4ZK/EgS/cQ=
 github.com/kr/logfmt v0.0.0-20140226030751-b84e30acd515/go.mod h1:+0opPa2QZZtGFBFZlji/RkVcI2GknAs/DXo4wKdlNEc=
-github.com/kr/pretty v0.1.0 h1:L/CwN0zerZDmRFUapSPitk6f+Q3+0za1rQkzVuMiMFI=
 github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORNo=
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
-github.com/kr/text v0.1.0 h1:45sCR5RtlFHMR4UwH9sdQ5TC8v0qDQCHnXt+kaKSTVE=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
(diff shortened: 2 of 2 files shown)


wsl-builder/common/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/ubuntu/WSL-DsitroLauncher/wsl-builder/common
```
didrocks, line 1: This is more than a FYI than something that needs an immediate action:
In Go, we generally avoid package and modules that are named `common` or `utils`. This was the name of the directory ofc with the symlink hack to avoid creating separate modules, but now that it’s a proper module, I’m a little bit more annoyed (yes, that’s a nitpick) :)

However, I don’t find a better naming. Something like `release` will make us then rename any structs/methods to avoid stuttering as they will be `release.` For instance `ReleaseInfo` -> `Info` (as `Release.Info`).

But there is GetPath will end up here while not being too close to the "Release" package.

So, let’s keep it that way, (told you it’s a FYI in general in Go), apart if you can come up with a better name :)

wsl-builder/common/paths.go
```diff
@@ -9,7 +9,7 @@ import (
 
 // getPathWith return the parent directory containing subdirectory. It goes up from
```
didrocks, line 10: You need to rename the documentation accordingly `GetPathWith` (and fix my typo `return` -> `returns` as the form should be `ExportedSymbol <verb> …`)

wsl-builder/common/releasesinfo.go
```diff
@@ -9,7 +9,7 @@ import (
 	"time"
 )
 
-type wslReleaseInfo struct {
+type WslReleaseInfo struct {
```
didrocks, line 12: I found the reason why you didn’t see any complain about exported fields: `golint` repository [has been archived](https://github.com/golang/lint) and VSCode extensions seem to (I didn’t find it in the changelog) have disabled it recently.

I think we will use golang-ci linters with a set of default rules similar to the one we have in adsys (switching from `golint` to `revive` thus), but the story won’t be good as gopls would complain it’s not run in project root.

Anyway, documentation should be in format of `ExportedSymbol <verb> …`. As we have been able to supports this in our projects, let’s keep it that way. The same rule goes for any exported field in the struct below (if you need more explanation on some of them, do not hesitate to ask)

wsl-builder/prepare-assets/go.mod
```diff
@@ -4,7 +4,9 @@ go 1.16
...
 require (
 	github.com/spf13/cobra v1.1.3
-	github.com/termie/go-shutil v0.0.0-20140729215957-bcacb06fecae // indirect
+	github.com/termie/go-shutil v0.0.0-20140729215957-bcacb06fecae
+	github.com/ubuntu/WSL-DsitroLauncher/wsl-builder/common v0.0.0-00010101000000-000000000000
 	gopkg.in/gographics/imagick.v2 v2.6.0
-	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
+
+replace github.com/ubuntu/WSL-DsitroLauncher/wsl-builder/common => ../common
```
didrocks, line 12: Maybe add a comment to say "We always want to use the latest local version". Well done finding that out btw for someone not familar with Go, and I guess then, Go modules… :)

didrocks (changes requested):
This looks good, I think few changes are needed, but I’m very pleased for someone not familar with Go and its module on how far (and correct) you got with it :)

wsl-builder/prepare-assets/go.mod
CarlosNihelton: It bit me until I figured out how to fix. When the pain comes, we tend to find solutions. :smile: .
I'll improve the comments.

wsl-builder/common/go.mod
CarlosNihelton: BTW I didn't make everything public. Some functions were only used inside the package, so I didn't touch them. I think we should keep it that way until we find the need for them outside of the package. It is not meant to be used publicly outside of this project, so I think we will notice when that need arise. Is that OK for you?

wsl-builder/common/releasesinfo.go
```diff
@@ -9,21 +14,23 @@ import (
...
-	ShortVersion  string
-	IconVersion   string
-	ReservedNames []string
-
-	codeName    string
-	shouldBuild bool
+// WslReleaseInfo models information about the Ubuntu releases to be used
+// in the template files which are part of the WSL image and launcher build process.
+type WslReleaseInfo struct {
+	WslID         string   // uniquely identifies this distro in WSL context.
```
didrocks, line 20: The nitpick would be here: some comments starts with a capital letter, some not, and one is missing a leading space. Also, some don’t end up with a colon.

I’m sure there are other places when it’s not compliant in the code, so we can do a global cleanup in one shot.

didrocks (approved):

PR #47, opened 2021-10-19
Make launcher aware of Windows locale and current user name

CarlosNihelton:
The following improvements are subject to this pull request:
- Removing full command paths hardcoded in the C++ source to improve testability.
- Obtaining Windows locale and user name using Win32 API's. Matching with the Ubuntu Installer is done by:
    -  Real User Name => EXTENDED_NAME_FORMAT::NameDisplay
    -  User Name (unix name) => EXTENDED_NAME_FORMAT::NameSamCompatible (with the domain removed)
    
    _See [GetUserNameExW](https://docs.microsoft.com/en-us/windows/win32/api/secext/nf-secext-getusernameexw) and [EXTENDED_NAME_FORMAT](https://docs.microsoft.com/en-us/windows/win32/api/secext/ne-secext-extended_name_format) docs for further reference_
   
- Shared the above mentioned information with Ubuntu OOBE.
- Moved code related to OOBE to it's own file because of the growth imposed by the changes mentioned above. This affects the headers inside meta/* subfolders.


DistroLauncher/OOBE.cpp
```diff
@@ -0,0 +1,121 @@
...
+		std::wstring PreparePrefillInfo() {
+			std::wstring commandLine;
+			std::wstring prefillInfo = DistributionInfo::GetPrefillInfoInYaml();
+			if (!prefillInfo.empty()) {
+				// Write it to a file inside \\wsl$ distro filesystem.        
+				const std::wstring prefillFileNameDest = L"/var/tmp/prefill-system-setup.yaml";
+				const std::wstring wslPrefix = L"\\\\wsl$\\" + DistributionInfo::Name;
+				std::wofstream prefillFile;
+				// Glad that mixing slashes and backslashes that way is not a problem in Windows.
+				prefillFile.open(wslPrefix + prefillFileNameDest, std::ios::ate);
```
didrocks, line 39: Could we have any error on open/create that we can trap?

DistroLauncher/OOBE.cpp
```diff
@@ -0,0 +1,121 @@
...
+				prefillFile << prefillInfo;
+				prefillFile.close();
+				if (!prefillFile.fail()) // updating the command line if file was written.
```
didrocks, line 42: I would log a warning that prefilling failed.

DistroLauncher/OOBE.cpp
```diff
@@ -0,0 +1,121 @@
...
+					commandLine += L" --prefill=" + prefillFileNameDest;
+			}
+			return commandLine;
+		} // std::wstring DistributionInfo::PreparePrefillInfo().
```
didrocks, line 46: This is the flutter dev inside you :)

DistroLauncher/OOBE.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+
+namespace DistributionInfo {
+    // OOBE Experience.
+    ULONG OOBE();
+
+    // Generating a YAML string from Windows user and locale information.
+    std::wstring GetPrefillInfoInYaml();
+
+    static const wchar_t* OOBE_NAME = L"ubuntu_wsl_setup";
```
didrocks, line 28: Actually, the command line will be `wsl-setup` that will not be in the path on purpose (we want that all interactions to the OOBE is made from the launcher: setup or config, as the launcher will take the reboot/shutdown option that was passed and act accordingly).
It’s going to be installed in `/usr/lib/libexec`, and thus, we need an absolute path, which I think is OK in that case.

You can see it in https://github.com/ubuntu/wsl-setup (if you want, we can spend some time today/tomorrow for a basic presentation of packaging). This package will be available in jammy.
This helper mounts the needed snaps, mock a snap environment and chain it to the snap desktop helper which then (phew!) chain to the actual binary. This is what you pass as an argument: the desktop-installer binary or TUI from subiquity.

DistroLauncher/OOBE.cpp
```diff
@@ -0,0 +1,121 @@
...
+	{
+		ULONG uid = UID_INVALID;
+
+		// calling the oobe experience
+		DWORD exitCode;
+
+		// Prepare prefill information to send to the OOBE.
+		std::wstring prefillCLIPostFix = DistributionInfo::PreparePrefillInfo();
+
+		// Flutter installer is called ubuntu_wsl_setup in its repo. I'm assuming it will be found inside PATH
```
didrocks, line 61: Remember that it will be flutter installer or TUI experience. I think this is what Patrick is working on as part of the shutdown/reboot experience?

DistroLauncher/OOBE.cpp
```diff
@@ -0,0 +1,121 @@
...
+		}
+
+		// getting username from ouput
+		// Create a pipe to read the output of the launched process.
+		HANDLE readPipe;
+		HANDLE writePipe;
+		SECURITY_ATTRIBUTES sa{ sizeof(sa), nullptr, true };
+		if (CreatePipe(&readPipe, &writePipe, &sa, 0)) {
+			// Query the UID of the supplied username.
+			std::wstring command = L"cat /var/lib/ubuntu-wsl/assigned_account";
```
didrocks, line 81: Is that part still needed? I think subiquity is now filing wsl.conf with the correct user name directly, which should remove the need for this hack. Am I correct @patrick330602?

DistroLauncher/WindowsUserInfo.cpp
```diff
@@ -0,0 +1,112 @@
...
+//  On Debian systems, the complete text of the GNU General
+//  Public License version 3 can be found in "/usr/share/common-licenses/GPL-3".
+//
+
+// Holds together the user information retrieved from Win32 API's.
+
+#include "stdafx.h"
+
+namespace DistributionInfo {
+	// Keep this part private to this translation unit because I don't think they need to be exported.
```
didrocks, line 25: same remark towards the "I" :)

DistroLauncher/WindowsUserInfo.cpp
```diff
@@ -0,0 +1,112 @@
...
+			// this function should be changed to use a proper YAML manipulation library, such as yaml-cpp.
+			std::wstring toYaml() {
+				std::wstring fullYaml;
+
+				if (!localeName.empty()) {
+					fullYaml += L"Welcome:\n  lang: " + localeName;
+				}
+
+				if (!(realName.empty() && userName.empty())) {
+					fullYaml += L"\nWSLIdentity:";
```
didrocks, line 44: So, if we don’t have a local, the file will start with:
```

WSLIdentity:
```

1. I think you tested, but subiquity has no issue with this? (Standard yaml parsing shouldn’t, be better to be safe than sorry)
2. I think we can still inverse the logic:
  a. add \n to the end of localeName
  b. then starting directly with `WSLIdentity`

The file will then ends with a trailing newline, but this is already what you are doing if we have a realName or an userName. Also, there is currently no trailing newline if you only have a locale and no real/username.

DistroLauncher/WindowsUserInfo.cpp
```diff
@@ -0,0 +1,112 @@
...
+		{
+			DistributionInfo::WindowsUserInfo userInfo;
+			const int size = LOCALE_NAME_MAX_LENGTH;
+			WCHAR loc[size];
+
+			std::size_t result = GetUserDefaultLocaleName(loc, LOCALE_NAME_MAX_LENGTH); 
+			if (result > 0) {
+				std::wstring_view view{ loc,result };
+				std::size_t dashPos = view.find_first_of('-');
+				if (dashPos > 0 && dashPos < result) loc[dashPos] = '_';
```
didrocks, line 74: hum, maybe it’s just me, but can we have this in 2 lines with `{}` (the rest of the code does this)

DistroLauncher/WindowsUserInfo.cpp
```diff
@@ -0,0 +1,112 @@
...
+			}
+
+			WCHAR userRealName[size];
+			ULONG nSize = size;
+			BOOLEAN ret = GetUserNameEx(EXTENDED_NAME_FORMAT::NameDisplay, userRealName, &nSize);
+			if (ret != 0) {
+				userInfo.realName = std::wstring{ userRealName,nSize };
+			}
+			else {
+				DWORD error = GetLastError();
```
didrocks, line 86: Honest question: what do we do with the error?

didrocks (changes requested):
Great work! Yes, I have some comments, but as you can see, most are nitpicks, some are about logic though and some others are to prepare what we land (you can ignore it for now and we can replace the commands with the real ones once they land in distro/jammy).

Feel free to correct me if some of my comments don’t make sense for any reasons :)

PR #48, opened 2021-10-19
Prepare-build now can copy a local rootfs file.

CarlosNihelton:
With the proposed changes one now can `./prepare-build.exe prepare UbuntuPreview build.txt ~/Downloads/impish_exit_0_oobe.tar.gz::amd64` for example. That command will copy the file from that path supplied if valid, instead of downloading from some cloud, which improves testability by enabling developers to customize their images as they wish and ensure the WSL building workflow stays the same.

### DIFF
--- a/wsl-builder/prepare-build/build.go
+++ b/wsl-builder/prepare-build/build.go
@@ -31,6 +31,7 @@ var (
 	}
 )
 
+// prepareBuild finds the correct paths of the VS projects, prepare build assets and get rootfs images.
 func prepareBuild(buildIDPath, wslID, rootfses string, noChecksum bool, buildID int) error {
 	rootPath, err := common.GetPathWith("DistroLauncher-Appx")
 	if err != nil {
@@ -48,7 +49,7 @@ func prepareBuild(buildIDPath, wslID, rootfses string, noChecksum bool, buildID 
 		buildNumber = fmt.Sprintf("%d", buildID)
 	}
 
-	archs, err := downloadRootfses(rootPath, rootfses, noChecksum)
+	archs, err := getRootfses(rootPath, rootfses, noChecksum)
 	if err != nil {
 		return err
 	}
@@ -60,7 +61,7 @@ func prepareBuild(buildIDPath, wslID, rootfses string, noChecksum bool, buildID 
 	return nil
 }
 
-// extractAndStoreNextBuildNumber returns the next minor build number to use
+// extractAndStoreNextBuildNumber returns the next minor build number to use.
 func extractAndStoreNextBuildNumber(buildIDPath string) (buildNumber string, err error) {
 	defer func() {
 		if err != nil {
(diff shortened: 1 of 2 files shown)


wsl-builder/prepare-build/build.go
```diff
@@ -157,36 +157,62 @@ func downloadRootfses(rootPath string, rootfses string, noChecksum bool) ([]stri
 	return arches, g.Wait()
 }
 
-// downloadFile into dest
+// downloadFile into dest (or copy it if from a local path).
```
didrocks, line 160: I don’t really like that a downloadFile function can either copy or download. What do you think about moving this decision more on the caller side (or a function like `isLocalFile(uri string) bool` and deciding where to branch) ?

CarlosNihelton: The only one I can see seems not very elegant, replacing every call to downloadFile() by copyOrDownload(), which would be like:
```go
func copyOrDownload(url string, dest string) error {
	if isLocalFile(url){
		return copyLocalFile(url, dest)
	} 		
	
	return downloadFile(url, dest);
}
```
Would that be more idiomatic?

didrocks: Ok, I think in that case we should use a named function for closure we pass to `g.Go()`.
So, this would would become: `getRootfs (uri, rootPath, winArch string, needsChecksum bool) error` which would do:
```
if err := os.MkdirAll(winArch, 0755); err != nil {
    return err
}

if isLocalFile(url)  { // of maybe directly do the stat, if we keep all in the same function
    // probably add a warning that needsChecksum is not supported for local URI if true?
    return copyLocalFile(url, dest)
}

if err := downloadFile(rootfsURL, filepath.Join(rootPath, winArch, "install.tar.gz")); err != nil {
   return err
}

if !needsChecksum {
   return nil
}

u, err := url.Parse(rootfsURL)
if err != nil {
    return err
}
u.Path = filepath.Join(path.Dir(u.Path), "SHA256SUMS")
checksumURL := strings.ReplaceAll(u.String(), "%5C", "/")
checksumDest := filepath.Join(rootPath, winArch, "SHA256SUMS")
if err := downloadFile(checksumURL, checksumDest); err != nil {
    return err
}
if err := checksumMatches(filepath.Join(rootPath, winArch, "install.tar.gz"), filepath.Base(rootfsURL), checksumDest); err != nil {
    return err
}

return nil
```

(Note: I know downloadFile doesn’t follow this, but you don’t need to repeat the type name for arguments when they are the same in following arguments)

What do you think?

CarlosNihelton: Agreed. About the checksum, it could be supported but it is very unlikely that a dev or user will edit the SHA256SUMS file to add an entry matching her custom image. (I did that just for experimenting with the previous commit), but with the cost of having the function you proposed more complex, since we couldn't just `return copyLocalFile()` without checksum'ing.

So, I'll proceed with the warning and skip checksum in that case.

wsl-builder/prepare-build/build.go
```diff
@@ -157,36 +157,62 @@ func downloadRootfses(rootPath string, rootfses string, noChecksum bool) ([]stri
...
 	}
+}
 
+// private function created to avoid creating the destination file too early
+func copyToDestinationFile(source io.Reader, dest string) (err error) {
 	out, err := os.Create(dest)
 	if err != nil {
 		return err
 	}
+
```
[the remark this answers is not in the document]
CarlosNihelton: Agreed.

wsl-builder/prepare-build/build.go
```diff
@@ -157,36 +157,62 @@ func downloadRootfses(rootPath string, rootfses string, noChecksum bool) ([]stri
...
+		if resp.StatusCode >= 400 {
+			return fmt.Errorf("http request failed with code %d", resp.StatusCode)
+		}
+
+		return copyToDestinationFile(resp.Body, dest)
 	}
+}
 
+// private function created to avoid creating the destination file too early
```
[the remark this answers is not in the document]
CarlosNihelton: I added a few final dots here and there.

didrocks:
This looks exactly what I had in mind. Thanks a lot for doing this :+1:

PR #50, opened 2021-10-21
Extract OOBE related code into its own component

cnihelton:
Per [@patrick330602 's comments](https://github.com/ubuntu/WSL/pull/47#pullrequestreview-784081768) in PR #47, extracting the code related to the OOBE into it's component as a separate PR is desirable. That will be the foundation for the addition of the 'prefill' feature, in which launcher will acquire and pass to the OOBE user name and locale from Windows.
Also PR makes all Linux commands launched during the setup process relative to `$PATH`, which improves testability. I thought it was a too small change to make in it's own PR, that's why it's grouped here.

### DIFF
--- a/DistroLauncher/OOBE.h
+++ b/DistroLauncher/OOBE.h
@@ -29,5 +29,5 @@ namespace DistributionInfo {
     HRESULT OOBESetup();
 
     // OOBE executable name.
-    static TCHAR* OOBE_NAME = L"ubuntu_wsl_setup";
+    static TCHAR* OOBE_NAME = L"/usr/lib/libexec/wsl-setup";
 }


DistroLauncher/DistroLauncher.cpp
```diff
@@ -31,13 +31,15 @@ HRESULT InstallDistribution(bool createUser)
...
     DWORD exitCode;
-    hr = g_wslApi.WslLaunchInteractive(L"/bin/rm /etc/resolv.conf", true, &exitCode);
+    hr = g_wslApi.WslLaunchInteractive(L"rm /etc/resolv.conf", true, &exitCode);
     if (FAILED(hr)) {
         return hr;
     }
 
-    // doing stuff here
-	hr = g_wslApi.WslLaunchInteractive(L"/usr/bin/which /usr/bin/ubuntu-wsl-oobe", true, &exitCode);
+    // checking for the existance of the OOBE executable.
```
didrocks: minor typo but "existence"? :p

didrocks (commented):
This looks good to me, just a couple of things:
- how does it compare to https://github.com/microsoft/WSL-DistroLauncher/tree/master/DistroLauncher now? (as the goal is to rebase on this and have a minimal diff -- apart from extra files). Ofc, there is the absolute vs relative path, but this is ok.
- Oh, I just saw you first push to refresh meta/, unsure if you noticed but the CI autogeneration was PRing against your own branch to merge https://github.com/ubuntu/WSL/pull/49 which was doing this automatically for you (so, you could merge into your branch before the final merge).

DistroLauncher/DistroLauncher.cpp
CarlosNihelton: Indeed.

patrick330602 (approved):
Looks good to me

PR #59, opened 2021-11-12
DEENG-56 Assets templates updates

jibel:
Updated the templates of the assets for the LTS, the generic Ubuntu release and then Preview release.


didrocks:
New icons \o/

PR #61, opened 2021-11-16
Reduce image size

jibel:
Use pngquant to significantly reduce the image size.


didrocks:
Generated images looks good and way smaller, which should make the application building successfully again!

PR #63, opened 2021-11-18
Fallback OOBE into text mode if graphics are not supported

CarlosNihelton:
There is a --text parameter added to wsl-setup, which starts the TUI instead of the snap GUI. We need to add it dynamically from the launcher:
- if wslg is not installed
- or if current distro is using wsl 1

To achieve that, WSLApiLoader was modified to expose one hidden function `WslGetDistributionConfiguration()`, which can report the recently installed distro subsystem version, default environment variables etc. For this PR the only relevant piece of information is the subsystem version.

There could be other ways to find that information, like querying the Windows registry for the default system version. That idea in particular could lead into crashes if the OOBE is ran in the future in reconfigure mode, for instance, under a distro with the version modified to WSL 1.

The graphics capability in this proposal is detected indirectly by checking if the symlink `/tmp/.X11-unix` is resolved correctly and if the corresponding socket is open and listening. Proof of concept tests proved that the socket is open right after distro registration under WSL 2 with WSLg enabled.

### DIFF
--- a/DistroLauncher/OOBE.cpp
+++ b/DistroLauncher/OOBE.cpp
@@ -242,7 +242,6 @@ namespace DistributionInfo {
 			switch (value[0]) {
 			case L'1': // forced text mode.
 				return true;
-				break;
 			case L'2': // forced GUI mode, no autodetection.
 				return false;
 			case L'0':
--- a/DistroLauncher/OOBE.cpp
+++ b/DistroLauncher/OOBE.cpp
@@ -23,19 +23,19 @@ namespace DistributionInfo {
 		std::wstring PreparePrefillInfo();
 		HRESULT OOBEStatusHandling(std::wstring_view status);
 		bool EnsureStopped(unsigned int maxNoOfRetries);
-		static TCHAR* OOBE_NAME = L"/usr/libexec/wsl-setup";
+		const TCHAR* OOBE_NAME = L"/usr/libexec/wsl-setup";
 	}
 
-	bool DistributionInfo::isOOBEAvailable(){
+	bool isOOBEAvailable(){
 		DWORD exitCode=-1;
 		std::wstring whichCdm{ L"which " };
 		whichCdm.append(DistributionInfo::OOBE_NAME);
-		HRESULT hr = g_wslApi.WslLaunchInteractive(whichCdm.c_str(), true, &exitCode);
+		HRESULT hr = g_wslApi.WslLaunchInteractive(whichCdm.c_str(), TRUE, &exitCode);
 		// true if launching the process succeeds and `which` command returns 0.
 		return ((SUCCEEDED(hr)) && (exitCode == 0));
 	}
 
-	HRESULT DistributionInfo::OOBESetup()
+	HRESULT OOBESetup()
 	{
 		// Prepare prefill information to send to the OOBE.
 		std::wstring prefillCLIPostFix = DistributionInfo::PreparePrefillInfo();
@@ -46,12 +46,12 @@ namespace DistributionInfo {
 		}
 		// calling the OOBE.
 		DWORD exitCode=-1;
-		HRESULT hr = g_wslApi.WslLaunchInteractive(commandLine.c_str(), true, &exitCode);
+		HRESULT hr = g_wslApi.WslLaunchInteractive(commandLine.c_str(), TRUE, &exitCode);
 		if ((FAILED(hr)) || (exitCode != 0)) {
 			return hr;
 		}
 
-		hr = g_wslApi.WslLaunchInteractive(L"clear", true, &exitCode);
+		hr = g_wslApi.WslLaunchInteractive(L"clear", TRUE, &exitCode);
 		if (FAILED(hr)) {
 			return hr;
 		}
--- a/DistroLauncher/OOBE.cpp
+++ b/DistroLauncher/OOBE.cpp
@@ -23,6 +23,7 @@ namespace DistributionInfo {
 		std::wstring PreparePrefillInfo();
 		HRESULT OOBEStatusHandling(std::wstring_view status);
 		bool EnsureStopped(unsigned int maxNoOfRetries);
+		bool mustRunOOBEinTextMode();
 		const TCHAR* OOBE_NAME = L"/usr/libexec/wsl-setup";
 	}
 
@@ -40,8 +41,8 @@ namespace DistributionInfo {
 		// Prepare prefill information to send to the OOBE.
 		std::wstring prefillCLIPostFix = DistributionInfo::PreparePrefillInfo();
 		std::wstring commandLine = DistributionInfo::OOBE_NAME + prefillCLIPostFix;
-		// Fallback OOBE to text mode if graphics are not supported.
-		if (!Helpers::WslGraphicsSupported()) {
+		// OOBE runs GUI by default, unless arg --text is set.
+		if (mustRunOOBEinTextMode()) {
 			commandLine.append(L" --text");
 		}
 		// calling the OOBE.
@@ -169,7 +170,7 @@ namespace DistributionInfo {
 
 			// Before relaunching, give WSL some time to make sure 
 			// distro is stopped.
-			bool stopSuccess = EnsureStopped(30);
+			bool stopSuccess = EnsureStopped(30); // NOLINT(readability-magic-numbers): only used here.
 			if (!stopSuccess) {
 				// We could try again, but who knows why
 				// we failed to stop the distro in the first time.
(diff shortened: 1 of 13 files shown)


DistroLauncher/ProcessRunner.cpp
```diff
@@ -1,3 +1,22 @@
+//
+//    Copyright: 2021, Canonical Ltd.
+//  License: GPL-3
+//  This program is free software : you can redistribute itand /or modify
+//  it under the terms of the GNU General Public License as published by
+//  the Free Software Foundation, either version 3 of the License.
```
didrocks, line 6: Are you sure this is the standard GPL3 header? (the ", either version of 3" reads weirdly…)

Some snippets from Unity for instance is:
```
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License version 3 as
 * published by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.

```

DistroLauncher/WSLInfo.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+        return distributionVersion;
+    }
+
+    bool isWslgEnabled() {
+        return isX11UnixSocketMounted() && isWSLVDCPluginRegistered();
+    }
+
+    bool WslGraphicsSupported() {
+        // Could WSL 3 or greater exist in the future?
```
didrocks, line 79: good one, I think this is fine this way to assume > 1.

Btw, stupid C++ question, but we can’t compare directly to 1, in this PR and the previous one, I see you always cast it with 1u to unsigned, but the c++ compiler is not smart enough to do the casting at compile time as this is a const?

DistroLauncher/WSLInfo.cpp
```diff
@@ -0,0 +1,133 @@
...
+        return false;
+    }
+     
+/* =========================== INTERNALS ================================== */
+    namespace {
+        // Returns true if the WSLDVCPlugin.dll is found.
+        // That is an indication that WSLg is installed.
+        // Search paths are empirical and subject to change.
+        bool isWSLVDCPluginRegistered() {
+            //TODO: if this hack survives, find more possible locations
```
didrocks, line 94: Yeah, this is why this is a little bit fragile IMHO, do we really need it or can we just rely on the socket symlink to not be created by WSL directly?

didrocks (changes requested):
Good work!

I really wonder if we should keep the .dll probing, due to its fragility. Do you mind doing additional testing first to only check for the bindmounted socket?

Also, there is one part of the card which is not addressed:
```We need a switch on the launcher to try forcing one mode for wsl setup (--text and --gui for instance?)```
This is for testing purposes, so it can be done directly with an argument of the launcher exe or via an environment variable (maybe the env is better like `LAUNCHER_FORCE_GRAPHICS` (0 or empty auto detect, 1 force wsl, 2 force test? or any better naming :p)

File: DistroLauncher/ProcessRunner.cpp (hunk 1 above)
Comment (CarlosNihelton) [reply DistroLauncher/ProcessRunner.cpp:6]: Well, I should have asked instead... Before I started to add our functions in separate files not present in the upstream, the only file with a Canonical copyright in this repository was the python script `lp-distro-info`, which is copyrighted as GPL 3. So I just did Ctrl-C Ctrl-V 😅 .

Also, the upstream repository is MIT licensed.

What do you recommend for that?

Comment (didrocks) [reply DistroLauncher/ProcessRunner.cpp:6]: GPL-3 and blame @jibel for the incorrect template :p

File: DistroLauncher/WSLInfo.cpp (hunk 2 above)
Comment (CarlosNihelton) [reply DistroLauncher/WSLInfo.cpp:79]: Old habits. 😅 
Compiler is smart enough to figure it out. Technically, that is not a cast, but a way to avoid a cast. Writting the literal as 1u ensures compiler will represent the number 1 appropriatelly as a unsigned integer binary, instead of making it signed and casting later at the moment of comparison. That is not really necessary in simple situations like this one. I'll take it out.

File: DistroLauncher/ProcessRunner.cpp (hunk 1 above)
Comment (CarlosNihelton) [reply DistroLauncher/ProcessRunner.cpp:6]: 🤣 
So, copy-paste from Unity and make sure that python script will get fixed as well.

File: DistroLauncher/WSLInfo.cpp (hunk 3 above)
Comment (CarlosNihelton) [reply DistroLauncher/WSLInfo.cpp:94]: My reasonsoning was to try to ensure Windows will be able to show the apps via RDP. But since Windows is the actor who launches wslg system distro and bind-mounts that folder, Windows having done all that and not being able to establish the RDP session would represent an invalid state way out of our control. So, having the socket listening should be enough for our purposes, I guess. I'll take that out.

File: DistroLauncher/WSLInfo.cpp (hunk 2 above)
Comment (CarlosNihelton) [reply DistroLauncher/WSLInfo.cpp:73]: I took the opportunity of this readibility check and added a few more thru clang-tidy. There is now a .clang-tidy in DistroLauncher folder with a few checks enabled. Although Visual Studio integrates well with clang-tidy, it will catch a fair amount of warnings from upstream code which I believe we will not address so soon, so instead of making it part of the build system by default, I left it discretely laying on the project folder as an opt-in advisor. It worked quite nicely. I could find a couple of opportunities for improvement in our sources.

File: DistroLauncher/WSLInfo.cpp (hunk 3 above)
Comment (CarlosNihelton) [reply DistroLauncher/WSLInfo.cpp]: Done. `ls -l /tmp/.X11-unix` and `ss -lx | grep /tmp/.X11-unix/X0` both exiting 0 should be enough to believe WSLg is enabled.

File: DistroLauncher/WSLInfo.cpp (hunk 2 above)
Comment (didrocks) [reply DistroLauncher/WSLInfo.cpp:73]: Excellent! :)

File: DistroLauncher/WSLInfo.cpp (hunk 3 above)
Comment (didrocks) [reply DistroLauncher/WSLInfo.cpp]: :+1:

**Comment by didrocks** [comment]
Just did a quick review here: that looks perfect! The only missing part is the override autodetection one!

**Comment by CarlosNihelton** [comment]
> Good work!
> 
> I really wonder if we should keep the .dll probing, due to its fragility. Do you mind doing additional testing first to only check for the bindmounted socket?
> 
> Also, there is one part of the card which is not addressed: `We need a switch on the launcher to try forcing one mode for wsl setup (--text and --gui for instance?)` This is for testing purposes, so it can be done directly with an argument of the launcher exe or via an environment variable (maybe the env is better like `LAUNCHER_FORCE_GRAPHICS` (0 or empty auto detect, 1 force wsl, 2 force test? or any better naming :p)

I'm going to work on that now. I didn't consider the idea of testing. Indeed that will be a lot useful. The environment variable will result in 0 changes in upstream code, against the CLI option, which would cause us to either change main to inject the CLI arguments vector or change signature of the function `InstallDistribution()`.

Do you really think we need to force GUI during tests?

The way I see this is:

- Prefer GUI by default.
- Fall back to text if either the env var is set or graphics are not supported.

Thinking that way, I'd name it to `LAUNCHER_TEST_FORCE_TEXT`. If set, go to text mode. If 0 or empty, auto detect.

**Comment by CarlosNihelton** [comment]
> > Do you really think we need to force GUI during tests?
> 
> This was for distinguishing those cases in tests (before looking at the specifics):
> 
> * test failing because autodetection is broken
> * test failing because GUI is broken.
> 


I see your point. It does make sense. Let's call it  `LAUNCHER_FORCE_MODE` and let it be:
- 0 or unset = autodetection
- 1               = text mode
- 2               = GUI mode.

File: DistroLauncher/OOBE.cpp [hunk 4]
```diff
@@ -210,11 +211,45 @@ namespace DistributionInfo {
...
+			if (unset || notASingleChar) {
+				return !Helpers::WslGraphicsSupported();
+			}
+
+			// Expected result if the env var is correctly set:
+			// readResult == 1 && value[1] == NULL && value[0] in (0,1,2).
+			switch (value[0]) {
+			case L'1': // forced text mode.
+				return true;
+				break;
```
didrocks, line 245: nitpick: is it for the linter that you break here?

didrocks (approved):
One question because there had to be one. but otherwise, good, approving while you can clean before merging!

DistroLauncher/OOBE.cpp
CarlosNihelton: Nope! That was just me forgetting that one. 😅 Sorry about the silly mistake.

CarlosNihelton: Fixed. I'm going to merge it now.

PR #68, opened 2021-11-24
Misinterpreted distributionInfo var in WSLInfo.cpp

CarlosNihelton:
Learned from Microsoft that the proper way to detect WSL1 or WSL2 is by inspecting the 4th bit of distribution flags variable written by the function `WslGetDistributionConfiguration()`.
The expression `(1+((wslDistributionFlags & 0x08)>>3))` in the return statement is just a way to evaluate that bit and return 1 or 2.

### DIFF
--- a/DistroLauncher/WSLInfo.cpp
+++ b/DistroLauncher/WSLInfo.cpp
@@ -85,7 +85,8 @@ namespace Helpers {
         // One indication that the distro has graphical support enabled.
         bool isX11UnixSocketMounted() {
             const std::array cmds = {
-                L"ls -l /tmp/.X11-unix",
+                L"ls -l /mnt/wslg",
+                L"ls -l /tmp/.X11-unix | grep wslg",
                 L"ss -lx | grep \"/tmp/.X11-unix/X0\"",
             };
             // I'm sure is better to read this way than with the algorithm.
--- a/DistroLauncher/WSLInfo.cpp
+++ b/DistroLauncher/WSLInfo.cpp
@@ -85,10 +85,11 @@ namespace Helpers {
         // One indication that the distro has graphical support enabled.
         bool isX11UnixSocketMounted() {
             const std::array cmds = {
-                L"ls -l /mnt/wslg",
-                L"ls -l /tmp/.X11-unix | grep wslg",
-                L"ss -lx | grep \"/tmp/.X11-unix/X0\"",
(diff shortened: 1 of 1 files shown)


DistroLauncher/WSLInfo.cpp
```diff
@@ -81,7 +85,8 @@ namespace Helpers {
         // One indication that the distro has graphical support enabled.
         bool isX11UnixSocketMounted() {
             const std::array cmds = {
-                L"ls -l /tmp/.X11-unix",
+                L"ls -l /mnt/wslg",
```
didrocks, line 88: `/mnt` is a configurable target for mounting Windows drive. I think it’s the same for wslg (but need confirmation). I would thus not rely on that path.
However, we should check that /tmp/.X11-unix is not a dangling symlink.

didrocks (commented):
Happy to +1 once you fix the `/mnt` issue. The byte operation on the registry is good, thanks for the good comment next to it!

DistroLauncher/WSLInfo.cpp
CarlosNihelton: In that case, listing /mnt/wslg might be misleading. I'll do some tests just got the sake of learning, but I'll take that command out.

CarlosNihelton: You are absolutely right! If one changes the `[automount].root` path in `/etc/wsl.conf`, wslg gets mounted in the new path provided. The symlink `/tmp/.X11-unix` is adjusted as well.

CarlosNihelton: `ls -l /tmp/.X11-unix | grep wslg` doesn't prove the symlink is resolved. `ls -l /tmp/.X11-unix/` does.

didrocks: There are multiple ways to detect that the symlink we point to exists:
`readlink -e /tmp/.X11-unix` returns something (and exits 0) if the target exists, 1 otherwise without any output.

Jibel suggest also use -e in bash -c for instance. We can even find more creative ways for sure :)

CarlosNihelton: That also means we will have trouble to know when to stop thinking about that... 😅

CarlosNihelton:
Outdated.

PR #74, opened 2021-12-10
Clean up the shutdown feature

CarlosNihelton:
Hi @jibel !
With this pull request I aim to:

1. Remove external dependencies and management solutions I previously introduced per our meeting earlier this week.
2. Adopt a simpler solution for the launcher-command file left by the OOBE.
3. Introduce a way to add some automated testing.

The solution choosen for [2.] was a simple key-value pairs file, with a very strict grammar, parsed line by line. It is expected that this file will be very small, so reading it and parsing it in a blocking way should not be a problem. The parser function fails silently for each line not compliant with the expected grammar. This should give us some flexibility, but not enough to enable running arbitrary commands.

Testing is something bugging me for a while in this project, but the parser described above was somewhat involving, and code reviewing it might not be a lot pleasant. So knowing that there is a somewhat easy way to test it helps both myself and the reviewer. 😄 It is not yet the most beatiful way to test the sources because I made it in a separate solution, so that the DistroLauncher.sln is completely unaware of it. It works fine and only depends on msbuild command line to build.

My next goal is go back to the CI Q&A pull request, rebase it to remove any hack or reference to external packages and on top of it add the msbuild commands to build and run unit tests. With that we will have linting, formatting and automated testing in the CI. This project is yet in a state which is quite complicated to auto test it, but having this solution in place will motivate us (or myself at least) to decouple and add some coverage. We'll see how much we'll be willing to invest on that. I like the idea very much. Then I'll go back to Subiquity and fix the shutdown related code to generate the launcher-command file per described above.

Also, soon I'll be changing some naming conventions. I'd like to at least conclude the view I have in mind about the structure of this launcher. If you look at the imagebelow, you'll agree that most of it is already true. I just want to make it more explicit by adopting the conventions on the orange side of the image. I hope I'm not dreaming away and will be able to take some time to do it and increase testability of those sources as well. That is somewhat secondary. After Subiquity, the next priority will be the splash.

image


CarlosNihelton:
Ah! For now, while I don't integrate it to the CI, if you want to try out the auto tests by hand in the CLI, you can issue the following commands from the repository root directory:

```powershell
msbuild -p:RestorePackagesConfig=true -t:restore -p:Platform=x64 -p:Configuration=Release .\DistroLauncher-Tests\DistroLauncher-Tests.sln
msbuild -p:Platform=x64 -p:Configuration=Debug .\DistroLauncher-Tests\DistroLauncher-Tests.sln
.\DistroLauncher-Tests\x64\Debug\DistroLauncher-Tests.exe

```

That requires a somewhat updated version of MSBuild (anything above v16.5 should be fine according to the docs, although my tests were done in v16.11, shipped together with VS 2019 Community Edition).

DistroLauncher/WSLInfo.cpp
```diff
@@ -38,62 +44,97 @@ namespace Helpers {
...
+    WslInfo::~WslInfo()
+    {
+        //if (environtmenVars != nullptr && envVarsCount > 0) {
+        //    // discards the env variable string array otherwise they will leak.
+        //    for (int64_t i = envVarsCount - 1; i >= 0; --i) {
+        //        CoTaskMemFree(static_cast<LPVOID>(environtmenVars[i]));
+        //    }
+        //}
+    }
 } // namespace Helpers.
```
CarlosNihelton, line 140: I'm not happy with this class. That constructor should not even exist, and the attempt to cache those values acquired from WSL API in a short lived object doesn't make much sense. I guess this mission better fits in a wrapper for WSL API Loader. That would make somewhat easier to do some auto tests and let the cached values live for a longer time. I'll push some updates to this exact piece.

CarlosNihelton:
Attempting to make this smaller, due the constrained resources available for code review, I had to attain myself to the core of this PR which are the bullets 1 and 2 of the PR prelude, which were:

1. Remove external dependencies and management solutions I previously introduced per our meeting earlier this week.
2. Adopt a simpler solution for the launcher-command file left by the OOBE.

So, the solution for automated testing will be subject to another PR. I hope this update makes this pull request easier to digest. It is still a lot of code, but there are two hints:

1. You can completely skip the header file "DistroLauncher\expected.hpp". That was a dependency that I just added to our source tree for easy development. In other scenarios I would add this file to another folder or even make it a git submodule, but we attempt to not change any basic MSVC project configuration (other than adding source files).
2. The renaming from `YamlExtensions.h` to `OobeDefs.h` should be interpreted as deletion of one no-more-necessary file and creation of another one. I'm not quite sure why Git thinks I renamed the file and changed its whole content. Maybe I did it and don't remember. 😅

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+        launcherCmdFile.close();
+        // We don't want that file existing after actions were taken.
+        std::filesystem::remove(prefixedFilePath);
+
+    }
+
+    namespace
+    {
+        bool ensureDistroStopped(unsigned int maxNoOfRetries);
```
jpnurmi, line 71: a separate declaration is not necessary because the body is below and nothing is referencing this in between?

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+            VoidResult shutdownDistro()
+            {
+                const std::wstring shutdownCmd = L"wsl -t " + DistributionInfo::Name;
+                int cmdResult = _wsystem(shutdownCmd.c_str());
+                if (cmdResult != 0) {
+                    return nonstd::make_unexpected(std::runtime_error("Failed to invoke shutdown command."));
+                }
+
+                // Before relaunching, give WSL some time to make sure distro is stopped.
+                bool stopSuccess = ensureDistroStopped(30); // NOLINT(readability-magic-numbers): only used here.
```
jpnurmi, line 167: would be perhaps nice to define 30 as a constant at the top to make it easier to find/change later if necessary

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+            }
+
+            return VoidResult();
+        }
+
+        namespace Actions
+        {
+            VoidResult shutdownDistro()
+            {
+                const std::wstring shutdownCmd = L"wsl -t " + DistributionInfo::Name;
```
jpnurmi, line 160: is there a reason to allocate an std::wstring just to get an std::string below?

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+                }
+
+                // Returns true once we don't find the DistributionInfo::Name in process's stdout
+                auto output = runner.getStdOut();
+                if (output.find(DistributionInfo::Name) == std::wstring::npos) {
+                    return true;
+                }
+
+                // We don't need to be hard real time precise.
+                Sleep(997); // NOLINT(readability-magic-numbers): only used here.
```
jpnurmi, line 106: why exactly 997? :D

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+        // create enums or tables to work with that.
+        VoidResult config(const KeyValuePairs& launcherCmds)
+        {
+            auto configIt = launcherCmds.find("defaultUid");
+            if (configIt == launcherCmds.end()) {
+                // nothing to do
+                return VoidResult();
+            }
+            auto default_uid = std::any_cast<unsigned long>(configIt->second);
+            // TODO: Replace the default flags on this call by the current ones reading before writting.
```
jpnurmi, line 145: typo: writting

DistroLauncher/WindowsUserInfo.cpp
```diff
@@ -41,48 +41,34 @@ namespace DistributionInfo {
...
-                auto conv = Win32Utils::wide_string_to_utf8(localeName);
-                if (conv) {
-                    sections["Welcome"]["lang"] = conv.value();
-                } else {
-                    Helpers::PrintMessage(MSG_ERROR_CODE, HRESULT_FROM_WIN32(GetLastError()), conv.error().c_str());
+                fullYaml += L"Welcome:\n  lang: " + localeName + L'\n';
                 }
-            }
+
+            if (!(realName.empty() && userName.empty())) {
```
jpnurmi, line 50: it's a matter of taste but to me it would be easier to read as `!realName.empty || !username.empty()` :)

DistroLauncher/exit_status_parser.cpp
```diff
@@ -0,0 +1,79 @@
...
+            String
+        };
+        const std::map<std::string, SupportedTypes> grammar{{"action", SupportedTypes::String},
+                                                            {"defaultUid", SupportedTypes::UInt}};
+
+        nonstd::expected<KeyValuePairs, const wchar_t*> parseExitStatusFile(std::istream& file)
+        {
+            std::regex keyValueRe(R"(^\s*(\w+)\s*[=:]\s*(\w+).*$)",
+                                  std::regex_constants::icase | std::regex_constants::ECMAScript);
+            std::regex commentRe(R"(^\s*#+.*)", std::regex_constants::icase | std::regex_constants::ECMAScript);
```
jpnurmi, line 26: can these be constants?

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "stdafx.h"
+#include <errno.h>
```
jpnurmi, line 19: is <cerrno> preferred for c++?

jpnurmi (commented):
I have only some minor nitpick. Feel free to ignore if you disagree with any. :)

DistroLauncher/ExitStatus.cpp
CarlosNihelton: `shutdownCmd.c_str()` returns a `const wchar_t*`, which is what the `_wsystem()` function expects. The wstring allocation was to easily concatenate the command `wsl -t`with the `DistributionInfo::Name`. That could be done concatenating C-strings, it would be pretty much the same, just a little bit more complicated.

DistroLauncher/ExitStatus.cpp
CarlosNihelton: It is not. I made that because when ifstream::open fails it doesn't tell exactly why so I'd need to look into the result of the system call. But after reflecting over your question I realized that I have std::filesystem in C++17 (even used it downwards ... 😅 ). I can avoid file-not-found error easily. That's what I'll do. Thanks for pointing that out.

DistroLauncher/ExitStatus.cpp
CarlosNihelton: Indeed. In between moving functions around I ended up moving the calling one below the definition, thus this declaration is unnecessary. Thanks for pointing that out.

DistroLauncher/ExitStatus.cpp
CarlosNihelton: Just o be fancy and not promissing exactly one second delay in between checks. 😄

DistroLauncher/ExitStatus.cpp
CarlosNihelton: Good catch!

DistroLauncher/ExitStatus.cpp
CarlosNihelton: I was fighting against the linter on this. Your question makes the linter the winner. I'll declare a constant at top, but still inside the anonymous namespace.

DistroLauncher/WindowsUserInfo.cpp
CarlosNihelton: I'll go with the recommendation.

DistroLauncher/exit_status_parser.cpp
CarlosNihelton: Yes, they can. Good point.

DistroLauncher/ExitStatus.cpp
```diff
@@ -0,0 +1,203 @@
...
+            // File not found will be interpreted as there is nothing to do.
+            return;
+        }
+
+        auto parserResult = internal::parseExitStatusFile(launcherCmdFile);
+        if (parserResult.has_value()) {
+            // Finally take the actions and delete the file.
+            const KeyValuePairs& launcherCmds{parserResult.value()};
+            auto actionResult = act(launcherCmds);
+            auto configResult = config(launcherCmds);
```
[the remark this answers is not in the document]
CarlosNihelton: Indeed. There is no much that can be actually done for handling potential errors here, but at least I should print the error message.

DistroLauncher/ExitStatus.cpp
jpnurmi: Ok, that's fine. I've been totally spoiled by QString

jpnurmi (commented):
LGTM :+1:

PR #75, opened 2021-12-17
Test invoking clang-tidy without CMake

CarlosNihelton:
This results in a very big refactoring. Let's see how it goes.


CarlosNihelton:
That was supposed to go on my fork, not here. My bad.

PR #78, opened 2022-01-06
Upgraded SDK version to fix HCS_E_HYPERV_NOT_INSTALLED undeclared

CarlosNihelton:
Per conversation on https://github.com/microsoft/WSL-DistroLauncher/issues/103 the recommendation is to upgrade to the latest SDK version. The other option would be define the constant in our sources.

```cpp
#define HCS_E_HYPERV_NOT_INSTALLED _HRESULT_TYPEDEF_(0x80370102L)
```

Since GitHub documentation says this version of the SDK is supported on the Windows runners, I opted in for the first option. Should the upstream implement any other fix I'll revisit this topic, but for now this is enough our CI.


didrocks:
This looks good to me. There is no opened automated PR against this one meaning that you changed all the generated files correctly :p

Thanks for reaching upstream and let’s merge this!

PR #93, opened 2022-02-09
Console handling enhancements

CarlosNihelton:
While integrating the console service with the splash controller under development, it was noticed some gaps in the console handling:

- Hiding console wouldn't work in the Windows Terminal
- GetConsoleWindow() misleads if the app is not running in the old console
- The order of assignment of handles and file descriptors was hiding Linux apps stdout and stderr.
- Splash controller will need to know the console window handle at some point.
- Console service sometimes needs to disconnect the pipe before its destruction.
- Fake (test) pipes had to update their API.

This PR fixes all those situations.

### DIFF
--- a/DistroLauncher/console_service.h
+++ b/DistroLauncher/console_service.h
@@ -11,12 +11,12 @@ namespace Win32Utils
         HANDLE stdErrHandle = nullptr;
         // The identity of this state is embedded into the values of the handles.
         // File descriptor values will most likely be always new since they are created by calling _dup().
-        friend constexpr bool operator==(const ConsoleState& lhs, const ConsoleState& rhs);
+        friend bool operator==(const ConsoleState& lhs, const ConsoleState& rhs)
+        {
+            return lhs.stdErrHandle == rhs.stdErrHandle && lhs.stdOutHandle == rhs.stdOutHandle;
+        }
     };
-    constexpr bool operator==(const ConsoleState& lhs, const ConsoleState& rhs)
-    {
-        return lhs.stdErrHandle == rhs.stdErrHandle && lhs.stdOutHandle == rhs.stdOutHandle;
-    }
+
     /**
      * Offers the application the capability to redirect the console output and toggle visibitily of the console window.
      * The console redirection is dependent on a class conforming to the same API as the LocalNamedPipe (see
--- a/DistroLauncher/console_service.h
+++ b/DistroLauncher/console_service.h
@@ -11,11 +11,12 @@ namespace Win32Utils
         HANDLE stdErrHandle = nullptr;
         // The identity of this state is embedded into the values of the handles.
         // File descriptor values will most likely be always new since they are created by calling _dup().
-        friend bool operator==(const ConsoleState& lhs, const ConsoleState& rhs)
-        {
-            return lhs.stdErrHandle == rhs.stdErrHandle && lhs.stdOutHandle == rhs.stdOutHandle;
-        }
+        friend constexpr bool operator==(const ConsoleState& lhs, const ConsoleState& rhs);
     };
+    constexpr bool operator==(const ConsoleState& lhs, const ConsoleState& rhs)
+    {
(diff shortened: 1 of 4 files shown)


DistroLauncher/console_service.h
```diff
@@ -11,11 +11,12 @@ namespace Win32Utils
         HANDLE stdErrHandle = nullptr;
         // The identity of this state is embedded into the values of the handles.
         // File descriptor values will most likely be always new since they are created by calling _dup().
-        friend bool operator==(const ConsoleState& lhs, const ConsoleState& rhs)
-        {
-            return lhs.stdErrHandle == rhs.stdErrHandle && lhs.stdOutHandle == rhs.stdOutHandle;
-        }
+        friend constexpr bool operator==(const ConsoleState& lhs, const ConsoleState& rhs);
```
didrocks, line 14: I am not sure to understand this change?

DistroLauncher/console_service.h
```diff
@@ -63,17 +65,25 @@ namespace Win32Utils
...
         }
 
       public:
         // std::forward<Pipe>(pipe) is just a way to ensure pipe is moved instead of copied.
         ConsoleService(Pipe&& pipe) noexcept : redirectTo{std::forward<Pipe>(pipe)}
         {
             assert(redirectTo.readHandle() != nullptr);
+            // Attempting to find the console window is best-effort if it is not the old style console.
+            // This has been tested for the new Windows Terminal. Other terminals out there may not fit.
+            window_ = FindWindow(L"CASCADIA_HOSTING_WINDOW_CLASS", DistributionInfo::WindowTitle.c_str());
```
didrocks, line 83: hum, is this working with the default windows command line app? as ubuntu.exe will be launched by default with it.

DistroLauncher/console_service.h
```diff
@@ -63,17 +65,25 @@ namespace Win32Utils
...
+            if (window_ == NULL) {
+                window_ = GetConsoleWindow();
```
didrocks, line 85: I guess this is the function which finds back the default?

didrocks (approved):
Looks good to me, interested in the answers but I trust you and thus can approve already :)

DistroLauncher/console_service.h
CarlosNihelton: I tested in a Windows 10 VM launching by double click on the launcher exe and by invoking it from powershell and by the command prompt (both running on the old school console). It falls back nicely. But it has to be the last case, because as I said, GetConsoleWindow() always return something. Checking whether that's not the desired window will be the same amount of work to find the desired window anyway.

DistroLauncher/console_service.h
CarlosNihelton: It depends on user settings. In my laptop everything that launches a console effectively runs on the new Windows Terminal. I'd expect from the majority of the users to be the old console. Very, very few should run something other than those two options (my guess based on the how easy is to access those alternatives, I don't have data to sustain that).

DistroLauncher/console_service.h
CarlosNihelton: This was me shuffling around the code while investigating why sometimes the console restauration was not behaving well for Linux apps. The only difference is that this operator now allows comparison in compile time, but it is very unlikely that will ever make sense. So, in practice nothing changes. I should have dropped it, but I forgot.

CarlosNihelton: Removed now.

DistroLauncher/console_service.h
didrocks: Hum, I think I will let you decide. Indeed, ` U+200E` is too magic in my opinion. If the current approach is good enough, let’s keep it. If we see more corner cases, maybe we can reevaluate that? Anyway, that’s really up to you :)

CarlosNihelton: Alright. Thank you. I'll keep it off of the table for now.

PR #99, opened 2022-02-21
Terminal theme fixes

didrocks:
Terminal theme fix, with the platform number fix which pass also in CI.

Note that it can look weird that the json is under an "Image" item type. There is no defined name, but I tried "None", "Item" and a set of assets + others and they do not work.
Image does not seem to have any side-effects and works well. There is no official guide on the Windows Terminal example for now, so let’s use this for now.

This branch was both tested on CI (and package installed from CI) from https://github.com/ubuntu/WSL/actions/runs/1887626864 and with Visual Studio, local sideloaded build.


CarlosNihelton (approved):
LGTM

It looks really shiny. Just tested locally.
image

PR #102, opened 2022-02-23
Progress on download

didrocks:
Add progress on download to detect stuck connection on CI.

It seems that github -> cloud-images is not able to download anymore over http. However, the connection is successful but most of the time, really slow and stuck.
To see stuck connection, print regular progress. However, as we handle multiple downloads at the same time, we can’t erase and reprint on the current line.
Some examples of the output: https://github.com/ubuntu/WSL/actions/runs/1886596739

Used this opportunity to differentiate initial connection time to complete download time.

This is a run with progress: https://github.com/ubuntu/WSL/runs/5301754416


wsl-builder/prepare-build/build.go
```diff
@@ -224,19 +241,53 @@ func downloadFile(url, dest string) (err error) {
...
+}
+
+func (wc *WriteCounter) Close() error {
+	return wc.f.Close()
 }
 
 // writeContentInto writes the content of the `source io.Reader`
 // into a new file created at `dest`.
-func writeContentInto(source io.Reader, dest string) (err error) {
+func writeContentInto(source io.Reader, total uint64, dest string) (err error) {
```
CarlosNihelton, line 278: Doc comments should be updated to indicate the role of the new parameter

wsl-builder/prepare-build/build.go
```diff
@@ -224,19 +241,53 @@ func downloadFile(url, dest string) (err error) {
...
 
-	return writeContentInto(resp.Body, dest)
+	var size int
+	if size, err = strconv.Atoi(resp.Header.Get("Content-Length")); err != nil {
+		log.Printf("Warning: unknown size for %s: %v", url, err)
+	}
+	return writeContentInto(resp.Body, uint64(size), dest)
+}
+
+type WriteCounter struct {
```
CarlosNihelton, line 251: Consider documenting public types and functions.

CarlosNihelton (approved):
Just a minor nitpick on doc comments, otherwise this looks really shiny.

wsl-builder/prepare-build/build.go
didrocks: good one!

wsl-builder/prepare-build/build.go
didrocks: It should stay rather private, let me change it

didrocks:
everything pass. Let’s merge

PR #103, opened 2022-03-07
Fix on close

CarlosNihelton:
During tests it was noticed that deadlocks could occur if the user closes the splash window before the application restored the console. We need to have the operating system to notify the application code when the child process ends, if that ever happen, asynchronously. Using only raw Win32 API requires multithreading. 

The heart of the changes are the pair `RegisterWaitForSingleObject/UnregisterWait`. The first assigns a thread from the system's thread pool to execute a callback we supply whenever the monitored object becomes signaled (in this case the splash process handle being closed). As this will be executed in another thread, care must be taken in the callback itself to make the call thread safe. As this monitoring is demanded only by the case of unexpected close-out (most likely by users) `UnregisterWait` must be called before executing a programmatic splash window close, thus the CloseEvent has to carry a way to `do_unsubscribe`. The supplied callback is stored in a type-erased `std::function<void()>`, so it can be a stateless or statefull callable object.

To further harden this PR to eliminate the risk of deadlocks due lack of console reader, the dynamics between the splash and installer GUI were changed in a way that the splash will be preferrably hidden instead of closed when it's time to leave room for the installer window to show up. This was implemented on the splash side on [PR Nº13](https://github.com/ubuntu/ubuntu-wsl-splash/pull/13).

The last piece of change in this PR was just a correction on the SplashController destructor, since it was calling Win32 API directly, defeating the purpose of having the SplashStrategy class.

```diff
        ~SplashController()
        {
            if (splashCloseNotifier != nullptr) {
                // This ensures that the OS will not call us back when there is no controller alive anymore.
                Strategy::do_unsubscribe(splashCloseNotifier);
            }
            if (procInfo.hProcess != nullptr) {
-                TerminateProcess(procInfo.hProcess, 0);
-                CloseHandle(procInfo.hThread);
-                CloseHandle(procInfo.hProcess);
+                Strategy::do_cleanup_process(procInfo);
            }
        }
```

Yet, as before, this is still isolated to the main program and won't cause any change in the launcher application's behavior.

### DIFF
--- a/DistroLauncher/splash_controller.h
+++ b/DistroLauncher/splash_controller.h
@@ -155,28 +155,21 @@ namespace Oobe
 
         static bool do_hide_window(HWND window)
         {
-            return ShowWindow(window, SW_HIDE) != 0;
+            // This enables Flutter code to react on our hide request.
+            constexpr auto WM_CUSTOM_AUTO_HIDE = WM_USER + 7;
+            PostMessage(window, WM_CUSTOM_AUTO_HIDE, 0, 0);
+            return true;
         }
 
         static bool do_place_behind(HWND toBeFront, HWND toBeBehind)
         {
             return SetWindowPos(toBeBehind, toBeFront, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE) != 0;
         }
-        // In the visible state, closing the window should allow the user to confirm closing or doing some clean
-        // up. That is the semantic difference between WM_CLOSE (enables that behavior) versus WM_QUIT (closes
-        // the window unconditionally). Interestingly, a GUI application that implements that semantics (think
-        // of document editor showing a dialog asking if the user wants to save their changes) will hang forever
-        // if the window receives WM_CLOSE while hidden, because the OS will not bring it back to the light,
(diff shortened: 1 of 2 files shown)


DistroLauncher/splash_controller.h
```diff
@@ -211,18 +229,44 @@ namespace Oobe
...
 
       public:
         /// Initializes the control structures to later run the splash application. Accepts the splash executable
-        /// file system path and the HANDLE to the device that will act as the splash standard input.
-        SplashController(std::filesystem::path exePath, not_null<HANDLE> stdIn) : exePath{std::move(exePath)}
+        /// file system path and the HANDLE to the device that will act as the splash standard input and a callable
+        /// object [onClose] to be invoked when the user closes the splash window, if that ever happens.
+        /// IMPORTANT NOTE ON MULTITHREADING: [onClose] will be invoked by another thread, thus any precaution required
+        /// to ensure safe concurrent call must be taken by the [onClose] callable implementation itself and any
+        /// functions it calls inside its body. That's the reason for the horrible type names.
```
didrocks, line 263: horrible, is that an euphemism? :)

didrocks (approved):
Looks good to me, and indeed, a better workflow/API for the new dynamic.

Merging it, thanks!

PR #105, opened 2022-03-09
Remove acrylic from theme

didrocks:
Remove acrylic as discussed last week.

I did a rebuild of Ubuntu Preview, and when installing it on ubuntudev1, the expected color profile (generated with Preview) is selected, so nothing more to be done in this regard AFAIK.


CarlosNihelton (approved):
Great! Thanks for doing that.

PR #106, opened 2022-03-14
WSLInfo.cpp/h maintenance

CarlosNihelton:
Those files were selected to receive the systemd related functions. As they were, CI would generated formatting and linting warnings when pushing systemd related code.

This PR acts as a preparation, anticipating and avoiding those warnings. At the same time, some conceptual changes moving code from `Helpers` namespace (which should stay upstream-only) to `Oobe::internal`. One function was left off this code move to avoid touching other files. They will be treated at a proper time.

Yet, some behavior changes were done here:
- Corrected the distroVersion interpretation.
- Added a check for the $DISPLAY env var being set for Wslg detection.

### DIFF
--- a/DistroLauncher/WSLInfo.cpp
+++ b/DistroLauncher/WSLInfo.cpp
@@ -17,29 +17,39 @@
 
 #include "stdafx.h"
 
-namespace Helpers {
-    namespace {
+namespace Oobe::internal
+{
+    namespace
+    {
         bool isX11UnixSocketMounted();
     }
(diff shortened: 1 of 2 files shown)


didrocks (approved):
This looks good to me and I haven’t spotted anything hurting my eyes :)

Thanks for working on this! Let’s merge it once we release Ubuntu new applications. I’ll do it.

didrocks:
And let’s merge now!

PR #107, opened 2022-03-15
Experimental systemd support on the launcher side

CarlosNihelton:
It builds on top of the setuid program we are working on [wsl-setup PR#1](https://github.com/ubuntu/wsl-setup/pull/1) to wrap any command the user passes to the launcher.

Command wrapping takes effect if the `boot.command` entry is found in `/etc/wsl.conf` and its value matches the distro expectation.

A set of test cases documents the assumptions and experiments done in WSL in regards to how it parses the `/etc/wsl.conf` to minimize false positives, i.e., situations in which we would consider systemd support active but WSL would not have activated it. Of course, that's best effort and many unforeseen situations leading to false positives may exist. It seems not scalable to try to prevent all.

### DIFF
--- a/DistroLauncher-Tests/DistroLauncher-Tests/IniFindValueTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/IniFindValueTests.cpp
@@ -1,0 +1,116 @@
+#include "stdafx.h"
+#include "gtest/gtest.h"
+
+#include "ini_find_value.cpp"
+
+// The assertions below derive from real life experiments in WSL.
+namespace Oobe
+{
+    namespace internal
+    {
+
+        TEST(IniFindValueTests, GoodFileShouldPass)
+        {
+            std::wstring buffer{LR"(# This mimics the syntax of /etc/wsl.conf file with comments.
+[user]
+default = root
+[boot]
+command = /usr/libexec/wsl-systemd
+)"};
+            std::wstringstream fakeFile{buffer};
+            ASSERT_TRUE(ini_find_value(fakeFile, L"boot", L"command", L"/usr/libexec/wsl-systemd"));
+            ASSERT_TRUE(ini_find_value(fakeFile, L"user", L"default", L"root"));
+            ASSERT_FALSE(ini_find_value(fakeFile, L"automount", L"enabled", L"true"));
+        }
+
+        TEST(IniFindValueTests, SemicolonCommentsBreaksWSL)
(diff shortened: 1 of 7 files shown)


DistroLauncher/ini_find_value.cpp
```diff
@@ -0,0 +1,95 @@
...
+                splitter.clear();
+                splitter.str(line);
+                std::getline(splitter, currentKey, L'=');
+                std::getline(splitter, value, L'=');
+                trim(currentKey);
+                trim(value);
+
+                // The only possible way to return true from this function is finding the key=*valueContains* definition
+                // we are looking for inside the proper section.
+                if (currentSection.compare(section) == 0 && currentKey.compare(key) == 0 &&
```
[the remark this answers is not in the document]
CarlosNihelton: Indeed. I forgot that I passed the function parameters as `wstring_view`s, which can compare to wstrings without causing allocation.

DistroLauncher-Tests/DistroLauncher-Tests/IniFindValueTests.cpp
```diff
@@ -0,0 +1,116 @@
...
+            ASSERT_FALSE(ini_find_value(fakeFile, L"boot", L"command", L"/usr/libexec/wsl-systemd"));
+        }
+        TEST(IniFindValueTests, SectionsCanBeExtended)
+        {
+            std::wstring buffer{LR"([automount]
+enabled = true
+[boot]
+command = /usr/libexec/wsl-systemd
+[automount]
+mountFsTab = true
```
didrocks, line 109: Nice use cases :)

I was wondering if it worths adding some with empty section (no child value) and some with stripping spaces around =.

didrocks (approved):
Nice use cases and additional testing :)

I’m approving it, however, just before merging, I was wondering if it worths adding some with empty section (no child value) and some with stripping spaces around =. at least to cover trim() and remove_surrounding_brackets(). What do you think?

DistroLauncher-Tests/DistroLauncher-Tests/IniFindValueTests.cpp
CarlosNihelton: Good that you asked that. It made me explore situations with spaces inside the section name. That seems to be interpreted as syntax error, since the configuration following such situations are not applied.

Now I'm pushing updates fixing this issue.

PR #108, opened 2022-03-18
Updated new logos for Ubuntu Preview

didrocks:


didrocks:
\o/ Let’s see how it goes with preview

PR #112, opened 2022-03-21
Typo fix in the script path for isotracker notification.

sil2100:
Currently ISO Tracker notifications fail as there is a typo in the script path - apologies!


didrocks:
Thanks @sil2100!

PR #113, opened 2022-03-21
Revisited descriptions.

jibel:
Updated the descriptions for all the products with the difference
between versioned and unversioned release. Also added a specific
description for Ubuntu Preview.

Co-authored-by: Didier Roche <didrocks@ubuntu.com>


didrocks:
Yeah, this is some duplication but I think this is the best path forward. We may add other keywords or something else, so let’s use this mecanism rather than attempting doing something too generic.

PR #117, opened 2022-03-24
Rename go build projects with real github path

didrocks:
We are now using wsl/ instead of distrolauncher in the package name.

Co-authored-by: Jean-Baptiste Lallement <jean-baptiste@ubuntu.com>


didrocks:
Tests are passing, and this will unblock next PR

PR #122, opened 2022-03-24
Adding the Flutter splash project as part of the bundle

CarlosNihelton:
This PR aims to bring the Flutter splash screen as a submodule of the launcher and bundle its build artifacts into the generated appx bundle. This is done by:

- Adding the Flutter repository as a submodule;
- Making the CI fetch and build it;
- Make the Appx project automatically bundle the Flutter build artifacts. This is achieved by the `Directory.build.props` file added to the Appx project directory. To make it specific to the releases that feature the OOBE, the referred build system file is added to the `meta/generated` directories. They are not available in the build tree by default.
- Ensuring the availability of the VC++ runtime on the target host. Achieved by adding the PackageDependency declaration to the `MyDistro.appxmanifest` files. This requires testing on a host we can ensure the runtime will be missing.

The current form of this PR has some drawbacks that may be matter for future efficiency enhancements:
- Flutter build artifacts built for x64 are bundled inside ARM64 packages. Flutter does not currently support ARM64 Windows targets, so the launcher will not even try to make use of those artifacts. 
- Splash flutter app is built at every iteration of the build matrix. The main reason is the icon change between WSLID's. We use the same icon of the launcher.

### DIFF
--- a/.github/workflows/build-wsl.yaml
+++ b/.github/workflows/build-wsl.yaml
@@ -93,6 +93,7 @@ jobs:
           git clone https://github.com/ubuntu/wsl.git ${{ env.workDir }}
           cd ${{ env.workDir }}
           git checkout "${GITHUB_SHA}"
+          git submodule update --init --recursive
       - name: Checkout build-info
         shell: bash
         run: |
@@ -134,6 +135,31 @@ jobs:
           $pwd = ConvertTo-SecureString  '${{ secrets.CERTIFICATE_PASSWORD }}' -AsPlainText -Force
           Import-PfxCertificate -Password $pwd -CertStoreLocation Cert:LocalMachine\Trust -FilePath certificate\certificate.pfx
           Import-PfxCertificate -Password $pwd -CertStoreLocation Cert:CurrentUser\My -FilePath certificate\certificate.pfx
+      - name: Check if packaging the splash app is necessary
+        working-directory: ${{ env.workDir }}
+        shell: bash
+        run: |
+          BUILD_TREE_HASH=$(sha256sum DistroLauncher-Appx/Directory.build.props | cut -f1 -d' ')
+          META_TREE_HASH=$(sha256sum meta/Ubuntu18.04LTS/generated/DistroLauncher-Appx/Directory.build.props | cut -f1 -d' ')
+          if [ ${BUILD_TREE_HASH} != ${META_TREE_HASH} ]; then
+            echo "withFlutterSplash=true" >> $GITHUB_ENV
+          else
+            echo "Skipping Flutter steps due hashes comparing equal:"
+            echo "    Build tree:  ${BUILD_TREE_HASH}. Meta tree: ${META_TREE_HASH}"
+          fi
+      - uses: subosito/flutter-action@v2
+        if: ${{ env.withFlutterSplash == 'true' }}
+        with:
+          channel: stable
+      - name: Build the splash screen
+        working-directory: ${{ env.workDir }}
+        if: ${{ env.withFlutterSplash == 'true' }}
+        run: |
+          flutter config --enable-windows-desktop
+          cp "DistroLauncher/images/icon.ico" "ubuntu-wsl-splash/windows/runner/resources/app_icon.ico"
+          cd "ubuntu-wsl-splash"
+          flutter gen-l10n
+          flutter build windows --release
       - name: Build Bundle
         working-directory: ${{ env.workDir }}
         run:  msbuild .\DistroLauncher.sln /t:Build /m /nr:false /p:Configuration=Release /p:AppxBundle=Always /p:AppxBundlePlatforms="${{ env.AppxBundlePlatforms }}" /p:UapAppxPackageBuildMode=${{ env.UapAppxPackageBuildMode }} -verbosity:normal
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,4 @@
+[submodule "ubuntu-wsl-splash"]
+	path = ubuntu-wsl-splash
+	url = https://github.com/ubuntu/ubuntu-wsl-splash
+	branch = main
--- a/DistroLauncher-Appx/Directory.build.props
+++ b/DistroLauncher-Appx/Directory.build.props
@@ -1,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+        <FlutterSplashDir>..\ubuntu-wsl-splash\build\windows\runner\Release\</FlutterSplashDir>
+    </PropertyGroup>
+	<ItemGroup>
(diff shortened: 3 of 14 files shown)


CarlosNihelton:
Sample taken from the sideload bundle generated by [this successful run](https://github.com/ubuntu/WSL/actions/runs/2035204914), with the Flutter artifacts highlighted in both x64 and ARM64 appx.

image

image

CarlosNihelton:
With the latest merge, this requires a careful reassessment whether it's ready or not. Let me put as draft for now.

CarlosNihelton:
#138 wants to undo the changes that target specific releases. I didn't want to apply those across all releases, since the prior ones will not benefit from the OOBE or the slide show.

CarlosNihelton:
Flutter steps are skipped when building older releases:

image

While for the newers are not:

image

CarlosNihelton:
The UbuntuPreview sideload bundle genereted [in this run](https://github.com/ubuntu/WSL/actions/runs/2092052527) contains the artifacts generated by Flutter inside the app bundle as expected per the explanation above.

x64 APPX:

image

ARM64 APPX:

image

didrocks (changes requested):
Happy that you got the flutter project separated and shipped in the bundle, nice work!

So, I’m not going to approve this right away, I think there are some drawbacks:
> To make it specific to the releases that feature the OOBE, the referred build system file is added to the meta/generated directories.
 and as you noticed:
> https://github.com/ubuntu/WSL/pull/138

The idea here is to never change manually the `generated` directories, as they are… generated! Also, ideally, we would avoid shipping things in `meta/src` for the point releases, as it means, everytime we open a new release, we will need to copy them.

So basically, the "master template" (as being project root content) corresponds to the next versioned LTS. Then, we can override previous versions.

So, the `appxmanifest` can be directly changed in the project directly. I don’t think there are some impacts on declaring it as UPW for all versions, right?

Then, the remaining question is about `Directory.build.props`. We should (by the logic above), ship it in `DistroLauncher-Appx/`. Then, we can have 2 strategies:
* either ships it everywhere (even if not used in older releases)
* or for those releases, have an empty override file in `src`, like `meta/Ubuntu20.004LTS/src/DistroLauncher-Appx/Directory.build.props`. Then, the github action workflow logic to install and build the flutter app can be changed to compare it to an empty sha256 or to `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` which seems to be the "empty file" sha256… or you can directly compare in the workflow with a `hasfiles()` of a known empty file, which is even more robust.

What do you think?

CarlosNihelton:
> Happy that you got the flutter project separated and shipped in the bundle, nice work!
> 
> So, I’m not going to approve this right away, I think there are some drawbacks:
> 
> > To make it specific to the releases that feature the OOBE, the referred build system file is added to the meta/generated directories.
> > and as you noticed:
> > #138
> 
> The idea here is to never change manually the `generated` directories, as they are… generated! Also, ideally, we would avoid shipping things in `meta/src` for the point releases, as it means, everytime we open a new release, we will need to copy them.
> 
> So basically, the "master template" (as being project root content) corresponds to the next versioned LTS. Then, we can override previous versions.
> 
> So, the `appxmanifest` can be directly changed in the project directly. I don’t think there are some impacts on declaring it as UPW for all versions, right?
> 
> Then, the remaining question is about `Directory.build.props`. We should (by the logic above), ship it in `DistroLauncher-Appx/`. Then, we can have 2 strategies:
> 
> * either ships it everywhere (even if not used in older releases)
> * or for those releases, have an empty override file in `src`, like `meta/Ubuntu20.004LTS/src/DistroLauncher-Appx/Directory.build.props`. Then, the github action workflow logic to install and build the flutter app can be changed to compare it to an empty sha256 or to `e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` which seems to be the "empty file" sha256… or you can directly compare in the workflow with a `hasfiles()` of a known empty file, which is even more robust.
> 
> What do you think?

I like the second option better, honestly. The only caviat is that the file cannot be completely empty, it must have the root element (<Project/>) declaration. But since that content will be exactly the same for the releases that won't feature the slide show, the comparison strategy remains the same, only the SHA changes to `9907733AF732AB3F3A691E9C3C0D55FEE72FDAFDAC410D639D46899A3169342D`  (according to Powershell) for the file containing only:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" />

```

CarlosNihelton:
Updated runs for:

- [Ubuntu22.04LTS (Jammy)](https://github.com/ubuntu/WSL/actions/runs/2097396165)

image


- [Ubuntu20.04LTS](https://github.com/ubuntu/WSL/actions/runs/2097228050)

image


As expected the Jammy app bundle contains the Flutter artifacts while the Focal doesn't.

.github/workflows/build-wsl.yaml
```diff
@@ -134,6 +135,31 @@ jobs:
...
+      - name: Check if packaging the splash app is necessary
+        working-directory: ${{ env.workDir }}
+        shell: bash
+        run: |
+          BUILD_TREE_HASH=$(sha256sum DistroLauncher-Appx/Directory.build.props | cut -f1 -d' ')
+          META_TREE_HASH=$(sha256sum meta/Ubuntu18.04LTS/generated/DistroLauncher-Appx/Directory.build.props | cut -f1 -d' ')
+          if [ ${BUILD_TREE_HASH} != ${META_TREE_HASH} ]; then
+            echo "withFlutterSplash=true" >> $GITHUB_ENV
+          else
+            echo "Skipping Flutter steps due hashes comparing equal:"
```
didrocks, line 147: The message could be enhanced to be due to the result: "Directory.build.props not doing anything" or so.

.github/workflows/build-wsl.yaml
```diff
@@ -134,6 +135,31 @@ jobs:
           $pwd = ConvertTo-SecureString  '${{ secrets.CERTIFICATE_PASSWORD }}' -AsPlainText -Force
           Import-PfxCertificate -Password $pwd -CertStoreLocation Cert:LocalMachine\Trust -FilePath certificate\certificate.pfx
           Import-PfxCertificate -Password $pwd -CertStoreLocation Cert:CurrentUser\My -FilePath certificate\certificate.pfx
+      - name: Check if packaging the splash app is necessary
+        working-directory: ${{ env.workDir }}
+        shell: bash
+        run: |
+          BUILD_TREE_HASH=$(sha256sum DistroLauncher-Appx/Directory.build.props | cut -f1 -d' ')
+          META_TREE_HASH=$(sha256sum meta/Ubuntu18.04LTS/generated/DistroLauncher-Appx/Directory.build.props | cut -f1 -d' ')
```
didrocks, line 143: Rather than META_TREE_HASH, I would give it more sense, like "NOOP_HASH"

didrocks (approved):
Let’s merge this as this is behaving as expected now. Do you mind having a look at my comment about naming and doing another PR with those fixed?

Thanks!

PR #124, opened 2022-03-24
wsl$ versus wsl.localhost

CarlosNihelton:
Not sure why this is happening, but sometimes trying to access `\\wsl$\<DISTRO>` blocks forever while `\\wsl.localhost\<DISTRO>` works.

While no enlightment comes to this issue, let's workaround.

Although this PR looks huge, that's mainly due formatting. The file `OOBE.cpp` predates the current CI and would trigger a lot of warnings. For actual content review focus on commit fbada24edb1ddcffef99cd4946068eb32d77b0bc

Now all filesystem accesses thru WSL Service in the launcher is done with `\\wsl.locahost` path prefix.

### DIFF
--- a/DistroLauncher/ExitStatus.cpp
+++ b/DistroLauncher/ExitStatus.cpp
@@ -52,7 +52,7 @@ namespace Oobe
 
     void ExitStatusHandling()
     {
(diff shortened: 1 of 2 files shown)


jibel (approved):
LGTM.

PR #128, opened 2022-03-29
Separate build template files

jibel:
In order to make the project builable from Visual Studio for example, separate
template files from project files and use specifci template files for project
files that must use templates:
- Use specific template files for project files
- Create template files and untemplate project files
- Ignore prepare-* at the root of the tree


CarlosNihelton (approved):
I had to test it locally end to end to be sure. This is as clean as it can be. Great job!

didrocks:
TBH, I’m not a very big fan of this. If upstream changes anything, there is a high chance of us missing them or badly copying/pasting in the template.
Also, I would have prefer that the master to be "UbuntuPreview' rather than Ubuntu, as it’s probably the application you want to destroy/replace the most.

What about rather than using template syntax, having magic words that we use to replace, and this way, keeping a single file instead of 2 for the ones we replace?

jibel:
Yeah, that's an idea too. I didn't want to obliterate the template
mechanism which was working fine.

PR #131, opened 2022-03-30
Silent which

CarlosNihelton:
Besides the typo in the variable name, while checking for the existence of the OOBE, this function caused some undesired output to console with not much context or information. This PR silences it, since for this function, checking the exit code is enough.


jibel:
Why are you using WslLaunchInteractive() instead of WslLaunch() and redirect stdout to a null handle? It seems cleaner to me than doing the redirection in the shell of the subprocess.

jibel (approved):

PR #132, opened 2022-04-01
Add some additional debug information to notify-isotracker runs

sil2100:
We also run notify-isotracker in debug mode temporarily - this should be switched off after we confirm all is good.


jibel (approved):
looks fine. Thanks.

PR #133, opened 2022-04-01
Fix secrets naming.

sil2100:
+ disable the debug flag from the job runs!


jibel (approved):
Alright, it matches now. Thanks.

PR #137, opened 2022-04-04
Extended command line options.

CarlosNihelton:
With PR #123 , #125 and #129 we added a new command line option to the launcher: `--enable-installer`.

This PR presents a more flexible way to add new command line options while preserving the existing ones, including the upstream. It is designed to avoid interaction with the upstream "command line parsing" (a series of if-else clauses in `DistroLauncher.cpp`) by removing the "extended" options from the vector of arguments (which is passed by a mutating reference).

The parsing output is a variant of types representing the tasks the launcher is supposed to do plus a `std::monostate` to represent the not-our-business options, which must be handled by the upstream code. Think of `std::monostate` like a `nil` type for variants.

Shall we want to support more command lines arguments in the future, it's just a matter of creating one new subtype in this file  and a tryParse function mimicking the existing implementations.

You'll notice a contrieved usage of templates inside the cpp file. That doesn't leak to the interface. It's used because for most of the options we added at this time are parsed in the very same way, except for `--autoinstall`, varying only in the quantity and contents of the required strings.

There are better command line parsing libraries out there, like `boost::program_options` or [docopt.cpp](https://github.com/docopt/docopt.cpp), but we are still avoiding bringing dependencies to the launcher and also, this requirement on not disturbing the "upstream CLI parsing" would require some customization anyway.

This, as well as the `MSG_USAGE_EXTENDED` message will be brought together by my next PR.

### DIFF
--- a/DistroLauncher/extended_cli_parser.cpp
+++ b/DistroLauncher/extended_cli_parser.cpp
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2021 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#include "stdafx.h"
+#include "extended_cli_parser.h"
+
+namespace Oobe::internal
+{
+    // Currently all CLI argument parsing outputs can be constructed in the same way, except for the AutoInstall,
+    // which requires a file path, thus templates with an specialization for the AutoInstall case. Those templates are
+    // only visible and accessible in this file.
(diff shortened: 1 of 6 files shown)


CarlosNihelton:
Besides the tests contained in this PR, the intended usage of this command line parsing mechanism can be seen [in this branch](https://github.com/ubuntu/WSL/blob/deeng-179-application/DistroLauncher/Application.h).

You can also see the final interconnection [in the same branch](https://github.com/ubuntu/WSL/blob/deeng-179-application/DistroLauncher/DistroLauncher.cpp).

DistroLauncher/extended_cli_parser.cpp
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (C) 2021 Canonical Ltd
```
didrocks, line 2: 2022, should be done in a future update :)

DistroLauncher/messages.mc
```diff
@@ -95,3 +95,37 @@ Language=English
...
+    <no args>
+        Launches the user's default shell in the user's home directory.
+
+    install [options]
+        Install the distribution and do not launch the shell when complete.
+          --root
+              Do not create a user account and leave the default user set to root.
+          --autoinstall <AUTOINSTALL-FILE-PATH>
+              Reads information from an YAML file to automatically configure the distribution.
+          --enable-installer
```
didrocks, line 113: I think this should rather be something like `--gui` or `--oobe` as the command is `ubuntu.exe install <with something>`. Also, we should consider forcing or not TUI.

We should review all together the UX for this, but let’s not block on it for now.

jibel (approved):
Some thoughts for future enhancements in comments. Otherwise LGTM. Thanks!

didrocks (approved):

PR #142, opened 2022-04-11
Use back templates as main files

didrocks:
Switch textual template with our markers.
Those markers should help producing binaries directly from the root
directory without needing to copy templates somewhere else and maintain
them separately.
The start delimiter is UbuntuDev.
The end delimiter is ..Dev

This should produce readable names, while still exposing what each value
is referring to.
Note that version/build number still need to be worked out to the
project fully buildable.


didrocks:
We now have an UbuntuDev.<xxxx>.Dev application, which is fully functional. No need for separate templates!

I think we should hook up building the flutter part (and initializing the submodule) as part of the build process to make it easier with custom commands during the build.

CarlosNihelton (approved):
LGTM. While testing this I discovered how to at least emit a better error message when Flutter built artifacts are missing. Topic for the next future, though.

didrocks:
\o/

PR #146, opened 2022-04-12
Corrects substring matching on distro shutdown.

CarlosNihelton:
Without the line break if user has two running distros and the name of one of those is a substring of the other,
the algorithm fails to detect the substring distro had shutdown.

For instance, let's say you have `Ubuntu` and `Ubuntu20.04LTS` running and you want to shutdown `Ubuntu`. The existing logic would fail to detect that `Ubuntu` shutdown because it would always find a match on `Ubuntu22.04LTS`.


didrocks:
ooopsss, indeed, annoying, makes sense with \r, approving!

didrocks (approved):

PR #150, opened 2022-04-13
Reverts `--enable-installer` option into a more complete `--ui=` command line.

CarlosNihelton:
Notice that the `=` signal is mandatory and spaces around are not tolerated in the current implementation. That can be enhanced if necessary, but will require more cleverness.

There is a header file trick to redefine the `MSG_USAGE` to `MSG_USAGE_EXTENDED` for the releases that support the OOBE. The header file `extended_messages.h` does this redefinition. For the releases that won't offer the OOBE, there should be an empty version of this file in meta subtrees, so the workflow can do the selection work.

With this implementation there is a nesting semantics on whether the shell should be available in the end or not (mandated by the argument `install` in the argv[1] position) versus whether the installer should be GUI, TUI or none (the original experience).

There are some compile-time tricks here and there to do some array manipulation in order to avoid repeating ourselves with the expected command line strings (which would otherwise not be checked until runtime).

As a result of a previous discussion, `--ui=gui` forces GUI mode without any checks whether the environment support graphics. Autodetection is available if no `--ui=` option is passed. The semantics of the `install` option is preserved, i.e. if present the launcher will just install the distro and quit. No shell in the end. Autointall option is only available with the `install` option, with the assumption that an automated installation should quit on exit, instead of launching a shell for user interaction.

The diagrams below summarize the command line options affected by this pull request:

WSL Launcher Command Line Reference

WSL Launcher Command Line Reference (1)

### DIFF
--- a/DistroLauncher-Tests/DistroLauncher-Tests/ExtendedCliParserTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/ExtendedCliParserTests.cpp
@@ -60,7 +60,7 @@ namespace Oobe::internal
 
     TEST(ExtendedCliParserTests, InstallOobeWithShellTui)
     {
-        // launcher.exe install --ui=tui
+        // launcher.exe --ui=tui
         std::vector<std::wstring_view> args{ARG_EXT_INSTALLER_TUI};
         auto opts = parseExtendedOptions(args);
         ASSERT_TRUE(std::holds_alternative<InteractiveInstallShell<OobeTui>>(opts));
--- a/DistroLauncher-Tests/DistroLauncher-Tests/ExtendedCliParserTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/ExtendedCliParserTests.cpp
@@ -6,6 +6,7 @@ namespace Oobe::internal
 {
     TEST(ExtendedCliParserTests, AutoInstallGoodCLI)
     {
+        // launcher.exe install --autoinstall <autoinstallfile>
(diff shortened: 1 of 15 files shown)


didrocks (commented):
I know this is still a draft but I couldn’t resist commenting:

> My reasoning was that users may request GUI (by explicitely commanding --installer=gui) without their environment supporting it.

I would say then "so be it", we should not prevent the user trying, because maybe we are wrong in our detection mecanism and there are corner cases.

Let me phrase it differently. If I want to remove a file which does not exist, the command still tries to remove the file, and then, the error is "file does not exist" which is returned by the kernel.

Said differently, again, it’s better in a program to do an action and check for errors than prevently do a check/capability existence (this avoid races between the check and the action and so on…).

CarlosNihelton:
> I know this is still a draft but I couldn’t resist commenting:
> 
> > My reasoning was that users may request GUI (by explicitely commanding --installer=gui) without their environment supporting it.
> 
> I would say then "so be it", we should not prevent the user trying, because maybe we are wrong in our detection mecanism and there are corner cases.
> 
> Let me phrase it differently. If I want to remove a file which does not exist, the command still tries to remove the file, and then, the error is "file does not exist" which is returned by the kernel.
> 
> Said differently, again, it’s better in a program to do an action and check for errors than prevently do a check/capability existence (this avoid races between the check and the action and so on…).

Got it. I'll proceed that way, then.

didrocks (approved):
This looks gorgeous, approving and let’s merge! Nice work :)

PR #151, opened 2022-04-14
Replace \\wsl$ or \\wsl.localhost path prefix with a function call

CarlosNihelton:
The result of the function call depends on the operating system version, which is obtained by reading the build number from the registry.

To avoid linting complains, some reformattings were done in `ExitStatus.cpp`. Since we are touching that file, the function `ExitStatusHandling()` was slightly simplified since the experimentation has shown that we can use wide characters in file paths to open files in the Linux file system. To simplify it further, I unified the return type of the error handling code into one single kind of exception to avoid slicing when returning errors.

### DIFF
--- a/DistroLauncher/ExitStatus.cpp
+++ b/DistroLauncher/ExitStatus.cpp
@@ -45,21 +45,15 @@ namespace Oobe
 
     namespace
     {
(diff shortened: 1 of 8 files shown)


didrocks (approved):
LGTM, tests pass, let’s merge to get it in today’s build!

PR #152, opened 2022-04-19
Fix build error due Mode undeclared for ARM64

CarlosNihelton:
The `using Mode = InstallerController<>::Mode;` statement was available only for x64. My bad. This fixes it and a related mistake in the cpp file as well.


didrocks:
tests pass, looks good to me, let’s go!

didrocks (approved):

PR #154, opened 2022-04-19
Makes use of more specific HRESULTS in installer_controller

CarlosNihelton:
And ensures the HRESULT value returning from a failed operation is preserved. Also increases the overall duration of the function  `poll_success()` to decrease he likelyhood of it causing an OOBE crash on slower machines.


didrocks:
I started to review before you assign it to me :) Do as the description says. Thanks! Merging :)

didrocks (approved):

PR #155, opened 2022-04-19
Really fix it, including the "." we replace now

didrocks:


CarlosNihelton (approved):
Now it's good not only for me :D

PR #157, opened 2022-04-19
Demote userinfo errors

CarlosNihelton:
Most of this PR is reformattings to avoid unrelated linting complaints.

The important point of it is the following change:

```diff
        inline void PrintLastError()
        {
            HRESULT error = HRESULT_FROM_WIN32(GetLastError());
-            Helpers::PrintErrorMessage(error);
+            wprintf(L"Recovering user information failed with the code: %lx\n", error);
        }
```

The rest is details.

This avoids the message to appear with the "Error: " prefix, thus reducing its level of importance. In fact, this is used for when GetUserNameEx and GetUserLocale functions fail, which is not a big thing.

Since this call `BOOLEAN ret = GetUserNameExW(EXTENDED_NAME_FORMAT::NameDisplay, userRealName, &nSize);` is the offeding one due the VM user not being fully setup in a domain with its detailed name registered, a second attempt is done by trying to read `NameGivenName` and `NameSurname`. Not a big fix, though.

image

The code was becoming messy with nested-if clauses, so I decided to take this out in a new function called `GetUserDisplayName()` and to preserve the symetry, I extracted a `GetUserRealName()` as well.

### DIFF
--- a/DistroLauncher/WindowsUserInfo.cpp
+++ b/DistroLauncher/WindowsUserInfo.cpp
@@ -17,11 +17,14 @@
 
 #include "stdafx.h"
 
-namespace DistributionInfo {
+namespace DistributionInfo
+{
     // Implementation details.
-    namespace {
+    namespace
+    {
         // WindowsUserInfo holds together the user information retrieved from Win32 API's.
-        struct WindowsUserInfo {
+        struct WindowsUserInfo
+        {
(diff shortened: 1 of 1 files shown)


didrocks (approved):
No objection on this one. Thanks for keeping the reformatting in its own commit :)

PR #161, opened 2022-04-20
Ensures OOBE will be presented on top on Reconfig.

CarlosNihelton:
Previously OOBE GUI placement was controlled by WSLG only. Sometimes it appeared behind other random windows.
With this change we leverage the same mechanisms used for interactive install to ensure the OOBE GUI will appear on top.
That requires breaking the reconfig in more states if GUI mode, reusing the same states related to the interactive install, 
maximizing reuse of the state machine components.

TUI reconfig remains unchanged.

Tests and comments were revised to reflect those changes as well.


didrocks (approved):
ack, this makes sense to leverage the same mecanism. We can merge it as we are going to redo a full test pass then. Good work :)

PR #162, opened 2022-04-21
Never launch OOBE in current working directory

CarlosNihelton:
Current working directory permissions may affect Subiquity startup or cause crashes.
OOBE should never be launched in cwd.


didrocks (approved):
Ack, indeed, better to be independent of any directory! Good :)

PR #168, opened 2022-04-25
Keep last store visibility state

didrocks:
This prevents making every new update for people only having the link.
We keep previous app state.


CarlosNihelton (approved):
LGTM

PR #169, opened 2022-04-25
Avoid building CI matrix when there is no distro to build

didrocks:
This ensure that we don’t have Includes: [] which is not supported by
GitHub Actions: "At least one vector element is required".


CarlosNihelton (approved):
LGTM

PR #181, opened 2022-04-26
Prefers server-status over OOBE exit code

### LINKED ISSUE #180: Installer reported failure but user was created
Graphical installer apparently completes its steps but the console shows error `0x80004004 Operation aborted` and applies the fallback method, which fails if the username provided is the same previously provided to the GUI because the installer did create the user in the first place.

# wsl --status
```
Distribuição Padrão: UbuntuDev.WslID.Dev
Versão Padrão: 2
WSL1 não é compatível com a configuração atual da máquina.
 Habilite o componente opcional "Windows Subsystem for Linux" para usar WSL1.

Versão do WSL: 0.58.3.0
Versão do kernel: 5.10.102.1
Versão do WSLg: 1.0.33
Versão do MSRDC: 1.2.2924
Versão do Direct3D: 1.601.0
Versão do Windows: 10.0.22598.200
```

# Windows build number

Windows 11 Dev channel (build 22598)

# Test Case
1. Make sure ubuntu-22.04 is not already registered
2. Run `ubuntu2204.exe --ui=gui` from command line (to ensure GUI mode will be applied).
3. Follow the graphical installation steps. Take a note of the username provided.
4. When allowed click on the "Finish" button.

## Actual result
Error message `0x80004004 Operation aborted` is displayed, fallback method is applied and it subsequently fails if the same user name is provided because the user was already created by the installer GUI.

```
Installer did not complete successfully.
Error: 0x80004004 Operation aborted

Applying fallback method.
Please create a default UNIX user account. The username does not need to match your Windows username.
For more information visit: https://aka.ms/wslusers
Enter new UNIX username: carlo
adduser: The user `carlo' already exists.
Enter new UNIX username:
```

## Expected behavior
No error messages and Linux shell opened with the newly created user logged in.

CarlosNihelton:
Windows 11 build 22598 users reported GUI mode installation reported failed but the background tasks had succeeded.
Thus, when applying the fallback with the same username, that method also failed.
OOBE is returning a failed exit code, but Subiquity concluded its work.
This ensures to return 0 if server-state is successful.
Returns OOBE error code otherwise.

Yet, further investigation on the OOBE side is required to understand why it's reporting failure if the installation completed and what is the relationship between the latest Windows and WSL updates and the OOBE reporting failure.

Fixes #180

### DIFF
--- a/DistroLauncher/installer_policy.h
+++ b/DistroLauncher/installer_policy.h
@@ -153,6 +153,9 @@ namespace Oobe
                 GetExitCodeProcess(process, &exitCode);
             }
 
+            [[maybe_unused]] DWORD clearExitCode = 0;
+            g_wslApi.WslLaunchInteractive(L"clear", FALSE, &clearExitCode);
+
             return exitCode;
         }
 
--- a/DistroLauncher/installer_policy.h
+++ b/DistroLauncher/installer_policy.h
@@ -147,15 +147,12 @@ namespace Oobe
                 return exitCode;
             }
 
-            GetExitCodeProcess(process, &exitCode);
-            // The Flutter installer exits 0 on some crashes.
-            // And sometimes reports crash even though Subiquity has finished. This gets fixed in [UDI PR
-            // 786](https://github.com/canonical/ubuntu-desktop-installer/pull/786)
-            if (exitCode == 0) {
-                g_wslApi.WslLaunchInteractive(L"grep -E 'EXITED|DONE' /run/subiquity/server-state", FALSE, &exitCode);
-                DWORD clearExitCode;
-                g_wslApi.WslLaunchInteractive(L"clear", FALSE, &clearExitCode);
+            // Whether the OOBE exit code is 0 or not, what trully matters is whether Subiquity finished its job:
+            g_wslApi.WslLaunchInteractive(L"grep -E 'EXITED|DONE' /run/subiquity/server-state", FALSE, &exitCode);
+            if (exitCode != 0) {
+                GetExitCodeProcess(process, &exitCode);
             }
+
             return exitCode;
         }
 


didrocks (commented):
You didn’t explain why you removed this clear execution, will we get on the console some debug output with grep?

As this only impact people with latest dev Windows version, I think we can clear that out before requalifying a new image.

DistroLauncher/installer_policy.h
```diff
@@ -147,15 +147,12 @@ namespace Oobe
...
             }
 
-            GetExitCodeProcess(process, &exitCode);
-            // The Flutter installer exits 0 on some crashes.
-            // And sometimes reports crash even though Subiquity has finished. This gets fixed in [UDI PR
-            // 786](https://github.com/canonical/ubuntu-desktop-installer/pull/786)
-            if (exitCode == 0) {
-                g_wslApi.WslLaunchInteractive(L"grep -E 'EXITED|DONE' /run/subiquity/server-state", FALSE, &exitCode);
-                DWORD clearExitCode;
-                g_wslApi.WslLaunchInteractive(L"clear", FALSE, &clearExitCode);
```
didrocks, line 157: You didn’t explain why you removed this clear execution, will we get on the console some debug output with grep?

As this only impact people with latest dev Windows version, I think we can clear that out before requalifying a new image.

CarlosNihelton: That was a mistake indeed, as I was experimenting with the code, I removed more than needed. 

This clear command is bugging me a little bit because it unconditionally clears the user's screen without considering whether it was doing something before on the console that should not be disturbed. Ideally, that clearing should only happen for the text-mode installation. But that's something I'll still clear my mind on to find the better way to do and then it will be a topic for another PR, not this one.

PR #182, opened 2022-04-27
build-remote: A few changes that were needed for usage on cdimage

sil2100:
Those changes are basically the following:

- Add support for anonymous LP login - I suppose using `login_with()` might have some uses (like for private PPAs), but in general for normal operation we don't really need to be logged in. Especially when running without any credentials in automation (like cdimage!).
- Add support for HTTPS proxy specification - this was needed due to the firewall situation on ancientminister.
- Remove usage of 3.9+ `removeprefix()` - for getting rid of the "ppa:" prefix, let's have a non-3.9+ method of stripping by providing a helper function with the good-ol' startswith()+slicing. This was needed as we don't have `removeprefix()` on ancientminister yet.

### DIFF
--- a/build-remote
+++ b/build-remote
@@ -48,6 +48,13 @@ def set_logging(debugmode=False):
     logging.debug("Debug mode enabled")
 
 
+def strip_ppa(string):
+    """Helper function for Python pre-3.9 to strip the ppa: suffix"""
+    if string.startswith("ppa:"):
+        return string[4:]
+    return string
+
+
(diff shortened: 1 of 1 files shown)


CarlosNihelton:
I didn't spot any issues, but since I'm not fully keen into this integration with launchpad, I'd like to ask @didrocks to have a look over it.

didrocks (approved):
Indeed, this looks good to me, let’s merge it. Thanks @sil2100!

PR #183, opened 2022-04-27
Removes prefill feature error messages

### LINKED ISSUE #170: Recovery user information failed with code 80070534
The following message appears in the console of the slideshow during installation of the distro:
Recovery user information failed with code 80070534

I don't see any subsequent error related to this but it is worrying for a user.

=Test Case=
1. Make sure ubuntu-22.04 is not already registered
2. Run ubuntu-22.04.exe for the first time to start the slideshow
3. Open the console under the slideshow and wait.

==Actual result==
The message "Recovery user information failed with code 80070534" is displayed

==Expected Result==
No error message

Tested on Windows 11 21H2 with Ubuntu 22.04

CarlosNihelton:
Errors herein printed are harmless to users, they are interesting to developers only. Thus instead of being displayed to users, they should be logged in a debug channel, where users won't see them but developers can in case more severe issues get reported. Since we don't yet have a proper logging, this PR removes printing any error messages related to acquiring user name and locale information.

Fixes #170 .

Leaving a TODO to place those to a debug channel for when we land a proper logging mechanism.

### DIFF
--- a/DistroLauncher/WindowsUserInfo.cpp
+++ b/DistroLauncher/WindowsUserInfo.cpp
@@ -32,14 +32,6 @@ namespace DistributionInfo
             std::string toYamlUtf8() const;
         };
 
-        // PrintLastError converts the last error code from Win32 API's into
-        // an error message.
-        inline void PrintLastError()
-        {
-            HRESULT error = HRESULT_FROM_WIN32(GetLastError());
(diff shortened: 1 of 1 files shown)


didrocks (approved):
ack then! :)

PR #184, opened 2022-04-27
Hide console only if it's dedicated to the launcher process

### LINKED ISSUE #176: Windows terminal hidden on first run of ubuntu2204.exe
Windows 11 21H2 / Ubuntu 22.04

=Test Case=
1. Open Windows Terminal with 2 tabs
2. With Ubuntu 22.04 installed but not registered run: ubuntu2204.exe

=Expected result=
Setup starts and the terminal window is still available so I can continue working while it's installing

=Actual result=
The terminal windows is hidden and reappears at the end of installation.

CarlosNihelton:
This PR is somewhat opinionated, so I'd expect some pushback.

The current implementation hides the console window unconditionally, behavior which drove issue #176 .

Ideally the console window should be hidden if it was only opened for the solely purpose of invoking the launcher to install the distribution, which is the case when the user launches it from start menu or from the store, right after installing the app. It could also be launched from the `Run` dialog (I don't know a better name for that, the one we get when we hit Super+R). In that case the user could pass an arbitrary combination of command line arguments and still a console window would be opened dedicated to the launcher process.

In contrast, not even for the empty command line case we would want to hide the console window if the user is doing something else in another tab of the Windows terminal, for instance.

There is not a clear way to detect that. Nor could I find a Win32 API call that could satisfy my inquiry about how is the launcher being launched.

The method I chose for this is: detect the current position of the console screen buffer cursor. If it matches the origin, thus we have our own dedicated window and we can and should hide it. Otherwise, user was doing something else and we better not disturb her. 

This only works if no console output is done before invoking the constructor of the `SplashEnabledStrategy` member of the `Application` class. This is easily achievable because the app object is created early in `main()` and its first member to be initialized is the strategy, which in turn calls the function responsible for memoizing whether at that moment the console cursor was at `{0,0}` as its first duty during construction (`launchedInDedicatedConsole()`). This sequencing is ensured by the order in which the members are declared inside the classes `Application<>` and `SplashEnabledStrategy`.

```cpp

int wmain(int argc, wchar_t const *argv[])
{
    // Update the title bar of the console window.
    SetConsoleTitleW(DistributionInfo::WindowTitle.c_str());

    // Initialize a vector of arguments.
    std::vector<std::wstring_view> arguments;
    for (int index = 1; index < argc; index += 1) {
        arguments.push_back(argv[index]);
    }

    Oobe::Application<> app(arguments);
``` 

I'm writting that much because, although this works I don't know when in the future somebody well intentioned would stick a printf between those calls and ruin the algorithm. Since the function `launchedInDedicatedConsole()`memoizes its return value once called for the first time, we could drop at call to it as the first line of `main()`. It would look wierd and increase slightly the diff to upstream, so I avoided that path. I considered creating a static global constant to force run this function before main(), but then I think we cannot ensure not falling into the static initialization order fiasco, since I don't know for sure whether our static globals would be initialized before or after the console handles being assigned to our process by the operating system, required to check whether the console is at the origin or at some offset.

For now the alternative I'm considering is iterating through all opened process to find the launcher parent's name and decide whether the console should be hidden or not based on that. Seems less performant and more fragile, thus not my first option.

I'm open to discussion and better ideas. Otherwise, this closes #176 .

### DIFF
--- a/DistroLauncher/extended_cli_parser.cpp
+++ b/DistroLauncher/extended_cli_parser.cpp
@@ -54,6 +54,11 @@ namespace Oobe::internal
 
     Opts parse(const std::vector<std::wstring_view>& arguments)
     {
+        // launcher.exe --hide-console - Windows Shell GUI invocation as declared in the appxmanifest. Hides the
+        // console, runs the OOBE (auto detect graphics support) and brings the shell at the end.
+        if (auto result = tryParse<ManifestMatchedInstall>(arguments); result.has_value()) {
+            return result.value();
+        }
         // launcher.exe - Runs the OOBE (auto detect graphics support) and brings the shell at the end.
         if (auto result = tryParse<InstallDefault>(arguments); result.has_value()) {
             return result.value();
--- a/DistroLauncher/extended_cli_parser.cpp
+++ b/DistroLauncher/extended_cli_parser.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Canonical Ltd
+ * Copyright (C) 2022 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as
--- a/DistroLauncher/extended_cli_parser.cpp
+++ b/DistroLauncher/extended_cli_parser.cpp
@@ -115,11 +115,20 @@ namespace Oobe::internal
     Opts parseExtendedOptions(std::vector<std::wstring_view>& arguments)
     {
         Opts options{parse(arguments)};
+
         // Erasing the extended command line options to avoid confusion in the upstream code.
         auto it = std::remove_if(arguments.begin(), arguments.end(), [](auto arg) {
             return std::find(allExtendedArgs.begin(), allExtendedArgs.end(), arg) != allExtendedArgs.end();
         });
         arguments.erase(it, arguments.end());
+
+#ifdef UNSUPPORTED_EXTENDED_CLI
+        if (shouldWarnUnsupported(options)) {
+            wprintf(L"Warning: ignoring command line options invalid for this Ubuntu release.\n");
+        }
+        return {};
+#else
         return options;
+#endif // UNSUPPORTED_EXTENDED_CLI
     }
 }
--- a/DistroLauncher/extended_cli_parser.h
+++ b/DistroLauncher/extended_cli_parser.h
@@ -84,10 +84,19 @@ namespace Oobe::internal
 
     // InteractiveInstallOnly<SkipInstaller> and InteractiveInstallShell<SkipInstaller> are actually std::monostate,
     // i.e. both go to upstream resolution.
-    using Opts = std::variant<std::monostate, AutoInstall, ManifestMatchedInstall, InstallDefault, InstallOnlyDefault,
-                              InteractiveInstallOnly<OobeGui>, InteractiveInstallOnly<OobeTui>,
-                              InteractiveInstallShell<OobeGui>, InteractiveInstallShell<OobeTui>, Reconfig>;
-
+    // clang-format off
+    // [clang-format appears to have trouble with long declarations like the following.]
+    using Opts = std::variant<std::monostate, // The upstream command line parsing effects.
+                              AutoInstall,
+                              ManifestMatchedInstall,
+                              InstallDefault,
+                              InstallOnlyDefault,
+                              InteractiveInstallOnly<OobeGui>,
+                              InteractiveInstallOnly<OobeTui>,
+                              InteractiveInstallShell<OobeGui>,
+                              InteractiveInstallShell<OobeTui>,
+                              Reconfig>;
+    // clang-format on
     /// Parses a vector of command line arguments according to the extended command line options declared above. The
     /// extended options are removed from the original vector as a side effect to avoid confusing the "upstream command
     /// line parsing" routines. Notice that argv[0], i.e. the program name, is presumed not to be part of the arguments
--- a/DistroLauncher/extended_cli_parser.h
+++ b/DistroLauncher/extended_cli_parser.h
@@ -93,4 +93,14 @@ namespace Oobe::internal
     /// line parsing" routines. Notice that argv[0], i.e. the program name, is presumed not to be part of the arguments
     /// vector. See DistroLauncher.cpp main() function.
     Opts parseExtendedOptions(std::vector<std::wstring_view>& arguments);
+
+    /// Returns true if the command line parsing result is invalid for the case when the extended CLI is unsupported
+    /// (the case for older releases).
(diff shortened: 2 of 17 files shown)


didrocks:
I think this will be better to discuss at a team meeting. I don’t really like the fact that it’s not straightfoward, and so, fragile which can lead to different user experience and issues.
Should we just not hide it then, and let it behind the splash screen + wslg window?

CarlosNihelton:
Digging into the AppxManifest documentation I learned about the Parameters option inside the UAP10 namespace, which allows specifying command line arguments with which the application should be launched. That serves almost perfectly to our use case and it is much more reliable, so I'm sticking with it. It affects launching the app from the Start Menu and from the store, but doesn't affect launching from command line, as expected. This PR now creates a new command line option to be used only in that context and all alone, not to be mixed with other existing command line options: `--from-start-menu-or-store`. If for curiosity user attempts to mix and match that parameter with pre-existing ones it may result in two scenarios:

1. During installation: Invalid command line arguments, fallback installation is applied and ends with a help message printed on the console.
2. During regular distro launching it is completely ignored.

To preserve the original experience with old releases I added specific versions of `MyDistro.appxmanifest` in the `meta/src` trees accordingly. 

I'm not sure about why #188 is changing all line endings on the generated counter parts of the above mentioned appxmanifest files, so I'll leave for @didrocks appreciation before merging that one onto this branch. I double checked whether the committed files in src preserved CRLF line endings.

CarlosNihelton:
Line endings solved. This is finally ready for review.

DistroLauncher/extended_cli_parser.cpp
```diff
@@ -54,6 +54,11 @@ namespace Oobe::internal
 
     Opts parse(const std::vector<std::wstring_view>& arguments)
     {
+        // launcher.exe --hide-console - Windows Shell GUI invocation as declared in the appxmanifest. Hides the
```
didrocks, line 57: mismatch: the option is named `--from-start-menu-or-store`.

TBH, I prefer `hide-console` which is what the option will do instead of basing on a policy based on where it starts from (and which can evolve)

meta/Ubuntu18.04LTS/src/DistroLauncher-Appx/MyDistro.appxmanifest
```diff
@@ -0,0 +1,43 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
didrocks, line 1: Why do we need an override there? It’s better to keep versioned distro with the root version (`[src/DistroLauncher-Appx/MyDistro.appxmanifest`. Let’s really have overrides when we need them.

meta/Ubuntu20.04LTS/src/DistroLauncher-Appx/MyDistro.appxmanifest
```diff
@@ -0,0 +1,43 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
didrocks, line 1: ditto

didrocks (changes requested):
So, went the flag path. I have some concerns on the name of the option itself.

Also, I emphase once again that we shouldn’t have a src/ in versioned directory and that their content should be what’s on root directories. This way, opening a new release does not force to copy any files or dirs.

DistroLauncher/extended_cli_parser.cpp
CarlosNihelton: That was the original idea, I changed my mind due the fact that later we may want to do other things besides hidding the console, but I guess you're right. I'm reverting to `--hide-console`.

meta/Ubuntu18.04LTS/src/DistroLauncher-Appx/MyDistro.appxmanifest
CarlosNihelton: My reasoning was: newer releases will behave as Ubuntu22.04 and Preview, which handle the new command line option to be invoked by the App Installer or Start Menu, but `18.04` and `20.04` don't have to, and because of that they would need their appxmanifest without the `uap10:Parameters` option. So I kept the previous version of the appxmanifest in the `meta/src` for those two releases while kept the root-tree manifest with the new behavior exactly to prevent the need to copy files as new releases appear.

---

I realized now (and just experimented with the idea) that we can avoid touching the build system if we let the code just ignore that option for the older releases. I can take advantage of the empty `extended_messages.h` files which exist to avoid leaking the new command line parsing behavior into the old releases to `#define` a preprocessor constant, let's say `UNSUPPORTED_EXTENDED_CLI`, and based on that constant disregard this and other extended command line options to preserve the original experience for the old releases. With this we won't need any manifest overrides anywhere.

I'm not a big fan of preprocessor macros, but this is a very simple and straightforward way to achieve what I'm looking for. I'm really tempted into pushing this idea forward.

How do you feel about that?

didrocks: Yes, if we were stable in terms of appxmanifest content, I wouldn’t have any issues in freezing it time older releases… So, I think your preprocessor approach would be better (and simply disable options we don’t want to work on those plateforms)

CarlosNihelton: Great, thanks! I'll revert the manifests override and do the preprocessor way.

DistroLauncher/extended_cli_parser.h
```diff
@@ -79,7 +84,7 @@ namespace Oobe::internal
 
     // InteractiveInstallOnly<SkipInstaller> and InteractiveInstallShell<SkipInstaller> are actually std::monostate,
     // i.e. both go to upstream resolution.
-    using Opts = std::variant<std::monostate, AutoInstall, InstallDefault, InstallOnlyDefault,
+    using Opts = std::variant<std::monostate, AutoInstall, ManifestMatchedInstall, InstallDefault, InstallOnlyDefault,
                               InteractiveInstallOnly<OobeGui>, InteractiveInstallOnly<OobeTui>,
                               InteractiveInstallShell<OobeGui>, InteractiveInstallShell<OobeTui>, Reconfig>;
```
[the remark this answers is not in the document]
CarlosNihelton: Resolved with c553545ba1761c1004ceb90fed3bfdf9473186f9

CarlosNihelton:
The current state:

- UAP10 namespace brought into the MyDistro.appxmanifest;
- uap10:Parameters option passing the command line option `--hide-console`;
- The change above affects launching from start menu or app installer, independent of which distro release;
- No appxmanifest overrides in the meta source trees;
- Command line parsing educated to warn about other options, but not this one;
- Console will only be hidden if the CLI matches exactly `launcher.exe --hide-console` and nothing else.

didrocks (approved):
Concerns addressed, thanks for fixing! Let’s merge :)

PR #200, opened 2022-05-13
Avoids CI break due Flutter upgrade

CarlosNihelton:
Some breaking changes were introduced with Flutter 3.0 that is causing the build workflow to break. The splash screen project was previously updated.
This updates the submodule


didrocks:
I guess you can automerge modules bump :) Doing it for you! :p

PR #203, opened 2022-05-26
Build windows app package on PR

didrocks:
Use for now a static rootfs, do not do any other change, and pull
everything to latest.
This ensures we can always build before merging our windows app package
and attach it to the PR to sideload, without any further change.

### DIFF
--- a/.github/workflows/build-pr.yaml
+++ b/.github/workflows/build-pr.yaml
@@ -1,0 +1,80 @@
+name: Build windows package
+on:
+  pull_request:
+
+jobs:
+  build-wsl:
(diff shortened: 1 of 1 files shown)


didrocks:
ok, build passes and we don’t have duplicated files with the last PR. We should check in the future on how we can store a given rootfs closer to CI so that we don’t rely on cloud-images download slowliness and still be able to test the final product.

CarlosNihelton (approved):
Thanks for doing this! I'll hold my wish to have a freezed smaller image to speed up the workflow.

PR #204, opened 2022-06-01
Enable github to jira bug synchronization

didrocks:
Use a default "jira" tag to import bugs to our Jira instance.


CarlosNihelton (approved):
Great work! Thanks for doing that!

PR #207, opened 2022-06-16
Added check around optional splash access

EduardGomezEscandell:
### Motivation
When using a rootfs corresponding to Ubuntu 20.04 LTS, a `std::bad_optional_access` exception is thrown.

### Changes
This is fixed by adding a check around it.


CarlosNihelton (approved):
LGTM! Thanks!

PR #208, opened 2022-06-16
Added exit status handling before launching

EduardGomezEscandell:
### Motivation
After upgrading Ubuntu Desktop, `ubuntu-release-upgrade` reboots the system.
In WSL this is not possible because there is no `systemd`.

### Solution
The package will set up `/run/launcher-command` to instruct the launcher to reboot on next launch.
As such, the launcher must now read this file before launching, and reboot if instructed so.
Otherwise, it continues as always.

### Status
Will stay in draft until I finish upgrading `ubuntu-release-upgrade`.


EduardGomezEscandell:
Closing to merge from a branch in the same repository (to allow usage of secrets).

PR #209, opened 2022-06-20
Fix no-return lambda warning

EduardGomezEscandell:
There is a lambda that we know always returns a value, but the compiler is unaware of this.

This fix adds a sanity test to ensure it trully never happens (in debug mode), and tells the compiler so in Release mode.

### DIFF
--- a/DistroLauncher/WSLInfo.cpp
+++ b/DistroLauncher/WSLInfo.cpp
@@ -160,6 +160,8 @@ namespace Oobe
             case Win32Utils::WinVersion::Win11:
                 return L"\\\\wsl.localhost\\";
             }
+            assert("Unsupported Windows version.", false);
+            __assume(0); // Unreachable
         }(Win32Utils::os_version());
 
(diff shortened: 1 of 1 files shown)


DistroLauncher/WSLInfo.cpp
```diff
@@ -157,9 +157,11 @@ namespace Oobe
             switch (version) {
             case Win32Utils::WinVersion::Win10:
                 return L"\\\\wsl$\\";
-            case Win32Utils::WinVersion::Win11:
-                return L"\\\\wsl.localhost\\";
+            /*case Win32Utils::WinVersion::Win11:
```
CarlosNihelton, line 160: Why commenting out the Win11 case?

EduardGomezEscandell: Oops, this was for debugging

CarlosNihelton (approved):
LGTM! Thanks!

PR #210, opened 2022-06-20
Enable launcher command after exiting

EduardGomezEscandell:
(Same as #208)

### Motivation
After upgrading Ubuntu Desktop, ubuntu-release-upgrade reboots the system.
In WSL this is not possible because there is no systemd.

### Solution
The package will set up /run/launcher-command to instruct the launcher to reboot, and advise the user to exit Ubuntu WSL.

Hence, the launcher must read this file after exiting Ubuntu WSL, and reboot if instructed so.
Otherwise, it does nothing.


CarlosNihelton (approved):
LGTM!

PR #218, opened 2022-07-11
Enables checking if localhost forwarding is disabled

CarlosNihelton:
That feature enables binding WSL 2 localhost to Windows localhost, thus enabling transparent connection through TCP where the client is running on the host while the server is running inside the WSL 2 distro, scenario expected by our OOBE. For security reasons, only localhost can connect to the server, thus this information is crucial.

That feature is enabled by default, but can be manually disabled by tweaking the `~\.wslconfig` file, which is a per-user configuration file that affects all distros running in the scope of the referred Windows user.

So to be able to read that configuration value we need to:

1. Find the user's home path;
2. Parse the file (ini parsing has been previously implemented)
3. look for the value of the entry [wsl2.localhostForwarding].

This will be used later in an upper level to determine whether we can run the OOBE on Windows or not. 

[Check the WSL docs](https://docs.microsoft.com/en-us/windows/wsl/wsl-config#:~:text=WSL%202%20VM.-,localhostForwarding,-boolean) for further information.

### DIFF
--- a/DistroLauncher-Tests/DistroLauncher-Tests/IniFindValueTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/IniFindValueTests.cpp
@@ -147,7 +147,33 @@ command = /usr/libexec/wsl-systemd
 
 )"};
             std::wstringstream fakeFile{buffer};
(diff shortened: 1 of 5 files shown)


EduardGomezEscandell (approved):
Makes sense to me

CarlosNihelton:
Just rebased to avoid a merge conflict due #220 .

PR #219, opened 2022-07-11
Prevents running the build workflow on draft PRs

CarlosNihelton:
That workflow is quite slow, due the need to download and package a real rootfs.
We need to figure out ways to make it faster, but in the meanwhile, let's skip it for draft PR's until they are marked as ready for review.

### DIFF
--- a/.github/workflows/build-pr.yaml
+++ b/.github/workflows/build-pr.yaml
@@ -1,11 +1,17 @@
 name: Build windows package
 on:
   pull_request:
+    types:
+        - opened
+        - reopened
+        - synchronize
+        - ready_for_review
 
 jobs:
   build-wsl:
     name: Build
     runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


CarlosNihelton:
The screenshot shows it disabling the worflow:
image

CarlosNihelton:
And now, after marking as ready for review:
image

.github/workflows/build-pr.yaml
```diff
@@ -1,11 +1,17 @@
...
+        - opened
+        - reopened
+        - synchronize
+        - ready_for_review
 
 jobs:
   build-wsl:
     name: Build
     runs-on: windows-latest
+    if: ${{ !github.event.pull_request.draft }}
```
EduardGomezEscandell, line 14: Will this prevent it from running in the main branch once merged?

EduardGomezEscandell (approved):
Great

PR #220, opened 2022-07-11
This app rootdir

CarlosNihelton:
Currently the splash executable (the slide show) is located inside the appx directory at the same level as the launcher. Soon we'll have the new OOBE located in the same directory, but a different filename. The patch feature being drafted in #217 will likely store files in a subdirectory as well. So, to avoid duplication, this PR extracts the logic to find this app root directory into it's own function, so we can append subdirectories and filenames as we wish.

The usage is illustrated in the `ApplicationStrategy.cpp` where we locate the `ubuntu_wsl_splash.exe` relative to that path.

### DIFF
--- a/DistroLauncher/ApplicationStrategy.cpp
+++ b/DistroLauncher/ApplicationStrategy.cpp
@@ -84,16 +84,7 @@ namespace Oobe
     {
         std::filesystem::path splashPath()
         {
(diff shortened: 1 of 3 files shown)


EduardGomezEscandell (approved):
Indeed this will be useful for #217.

Approved.

PR #221, opened 2022-07-13
Created class to deal with interprocess syncronization

EduardGomezEscandell:
Abstract
--------
A RAII class to deal with Windows inter-process synchronization.

Usage
--------
Best is to create a single NamedMutex as a static variable. If you believe it's not going to be used in most executions, it is a good idea to enable lazy initialization:
```c++
static auto my_mutex = NamedMutex("unique_identifyier", true);
```
Otherwise, use:
```c++
static auto my_mutex = NamedMutex("unique_identifyier");
```
These constructors obtain a handle to the Mutex, but do not acquire a lock. In order to do so, you have two options. Use the first one unless you've got a good reason not to.
 1. Use the monadic interface:
    ```c++
    my_mutex.lock()
      .and_then([]() {
          // do stuff
      })
      .or_else([] {
          // panic
      });
    ```
    Note that the ordering of the functions doesn't matter,
    and both are optional and repeatable.

 2. Use a scope guard:
    ```c++
    NamedMutex::Lock scope_guard = my_mutex.lock();
    if(scope_guard.ok()) {
        // do stuff
    } else {
        // panic
    }
    ```
    Be careful! If your program throws before the Guard goes out of scope, and it's not caught, your mutex may not be released. Option 1 takes care of this automatically, hence why it's recommended.

### DIFF
--- a/DistroLauncher/named_mutex.cpp
+++ b/DistroLauncher/named_mutex.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2022 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "stdafx.h"
+
+DWORD NamedMutex::create() noexcept
+{
+    mutex_handle = CreateMutex(NULL, FALSE, mutex_name.c_str());
+    if (!mutex_handle) {
+        const auto error = GetLastError();
+        return error ? error : ~0; // Ensuring return value is always != 0 when mutex_handle is nullptr
+    }
+    return 0;
+}
+
+DWORD NamedMutex::destroy() noexcept
+{
+    if (mutex_handle) {
+        return CloseHandle(mutex_handle);
+    }
+    return 0;
+}
+
+DWORD NamedMutex::wait_and_acquire() noexcept
+{
+    if (!mutex_handle) {
+        DWORD success = create();
+        if (success != 0) {
+            return success;
+        }
+    }
+    __assume(mutex_handle);
+    return WaitForSingleObject(mutex_handle, 1000);
+}
+
+DWORD NamedMutex::release() noexcept
+{
+    return ReleaseMutex(mutex_handle);
+}
--- a/DistroLauncher/named_mutex.h
+++ b/DistroLauncher/named_mutex.h
@@ -1,0 +1,159 @@
+/*
+ * Copyright (C) 2022 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+#pragma once
+
+#include "stdafx.h"
+
+/*
+ *  Abstract
+ *  --------
+ *  A RAII class to deal with Windows inter-process synchronization.
+ *
+ *  Usage
+ *  --------
+ *  Best is to create a single NamedMutex as a static variable. If you believe
+ *  it's not going to be used in most executions, then use
+ *  ```
+ *  static my_mutex = NamedMutex("unique_identifyier", true);
+ *  ```
+ *  to enable lazy initialization. Otherwise, use
+ *  ```
+ *  static my_mutex = NamedMutex("unique_identifyier");
+ *  ```
+ *
+ *  In order to acquire a lock, you have two options.
+ *  Use the first one unless you've got a good reason not to.
+ *   1. Use a lambda wrapper:
+ *      ```
+ *      DWORD response = my_mutex.Run([](){
+ *          // do stuff
+ *      });
+ *      if(response != 0) {
+ *          throw "Oh, no!";
+ *      }
+ *      ```
+ *
+ *   2. Use a scope guard:
+ *      ```
+ *      NamedMutex::Guard scope_guard = my_mutex.get();
+ *      if(scope_guard.ok()) {
+ *          // do stuff
+ *      } else {
+ *          throw "Oh, no!";
+ *      }
+ *      ```
+ *      Be careful! If your program throws and it's not caught,
+ *      your mutex may not be released. Option 1 takes care of
+ *      this automatically, hence why it's recommended.
+ */
+class NamedMutex
+{
+  public:
+    NamedMutex() = delete;
+    NamedMutex(NamedMutex&) = delete;
+    NamedMutex(NamedMutex&&) = delete;
+
+    /// Some mutexes are almost never used. This makes for an easy optimization by not initializing them unless needed.
+    NamedMutex(std::wstring name, bool lazy_init = false) :
+        mutex_handle(nullptr), mutex_name(L"WSL_" + DistributionInfo::Name + L"_" + name)
+    {
+        if (!lazy_init) {
+            create();
+        }
+    }
+
+    ~NamedMutex()
+    {
+        destroy();
+    }
+
+    class Guard
+    {
+      public:
+        constexpr Guard() noexcept : parent_(nullptr), response_(0)
+        { }
+
+        Guard(NamedMutex& parent) noexcept : parent_(&parent), response_(parent.wait_and_acquire())
+        { }
+
+        Guard(Guard& other) = delete;
+        Guard(Guard&& other) noexcept : Guard()
+        {
+            *this = Guard(std::move(other));
+        }
+
+        ~Guard() noexcept
+        {
+            if (ok()) {
+                parent_->release();
+            }
+        }
+
+        Guard& operator=(Guard&& other) noexcept
+        {
+            std::swap(parent_, other.parent_);
+            std::swap(response_, other.response_);
+            return *this;
+        }
+
+        bool ok() const noexcept
+        {
+            return parent_ && (response_ == WAIT_OBJECT_0);
+        }
+
+        DWORD why() const noexcept
+        {
+            return response_;
+        }
+
+      private:
+
+        NamedMutex* parent_;
+        DWORD response_;
+
+        friend class NamedMutex;
+    };
+
+    Guard get()
+    {
+        return Guard(*this);
+    }
+
+    template <typename Callable> DWORD Run(Callable f)
+    {
+        try {
+            if (DWORD failure = wait_and_acquire(); failure) {
+                return failure;
+            }
+            f();
+        } catch (std::exception except) {
+            release();
+            throw except;
+        }
+
+        return release();
+    }
+
+  private:
+    HANDLE mutex_handle;
+    std::wstring mutex_name;
+
+    DWORD create() noexcept;
+    DWORD destroy() noexcept;
+    DWORD wait_and_acquire() noexcept;
+    DWORD release() noexcept;
+};
--- a/DistroLauncher/DistroLauncher.vcxproj
+++ b/DistroLauncher/DistroLauncher.vcxproj
@@ -146,6 +146,7 @@
     <ClInclude Include="extended_messages.h" />
     <ClInclude Include="installer_controller.h" />
     <ClInclude Include="installer_policy.h" />
+    <ClInclude Include="named_mutex.h" />
     <ClInclude Include="not_null.h" />
     <ClInclude Include="OOBE.h" />
     <ClInclude Include="DistributionInfo.h" />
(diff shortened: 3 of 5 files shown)


DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,159 @@
...
+
+        ~Guard() noexcept
+        {
+            if (ok()) {
+                parent_->release();
+            }
+        }
+
+        Guard& operator=(Guard&& other) noexcept
+        {
```
CarlosNihelton, line 107: This seem suspecious to me: you're modifying both Guards by swapping their guts on an assignment operator call? If that's really what you want, it'd be better defining a free function called swap in the same namespace of the class.

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,159 @@
...
+        constexpr Guard() noexcept : parent_(nullptr), response_(0)
+        { }
+
+        Guard(NamedMutex& parent) noexcept : parent_(&parent), response_(parent.wait_and_acquire())
+        { }
+
+        Guard(Guard& other) = delete;
+        Guard(Guard&& other) noexcept : Guard()
+        {
+            *this = Guard(std::move(other));
```
CarlosNihelton, line 96: Is not this an infinite loop? I like to do `std::exchange`  to initialize the members in a move constructors, thus I leave the moved-from object in a known "empty" state.

DistroLauncher/named_mutex.h
EduardGomezEscandell: This is in case you want to pass the guard around. For instance:

```c++
struct Sudo {
NamedMutex::Guard sudo_guard (nullptr);

Sudo() {
   static auto root_user_mutex = NamedMutex("sudo", true);
   auto guard = root_user_mutex.lock();

   auto succeeded = SetUserAsRoot();
   
   if(succeeded) {
      this->sudo_guard = std::move(guard); // extending lifetime
   }
}

~Sudo() {
   ResetUser();
}

};

```

This way you can extend the guard lifetime outside its scope.

DistroLauncher/named_mutex.h
EduardGomezEscandell: No, this calls `Guard& operator=(Guard&& other)`.

EduardGomezEscandell: Agreed on the use of std::exchange though

EduardGomezEscandell: Actually it was an infinite loop. Fixed.

DistroLauncher/named_mutex.h
EduardGomezEscandell: Real example here:

https://github.com/ubuntu/WSL/blob/8f9f6a60a4f7af49245eb72c5246414ed778f9a8/DistroLauncher/sudo.h#L111-L133

DistroLauncher/named_mutex.cpp
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "stdafx.h"
+
+DWORD Win32MutexApi::create(HANDLE& mutex_handle, LPCWSTR mutex_name) noexcept
+{
+    mutex_handle = CreateMutex(nullptr, FALSE, mutex_name);
+    if (mutex_handle == nullptr) {
```
CarlosNihelton, line 23: Just a side comment: This is correct, but some Win32 functions returns `INVALID_HANDLE_VALUE` (`#define INVALID_HANDLE_VALUE ((HANDLE)(LONG_PTR)-1)`) instead of `NULL`, so make sure to always check the docs, because that value compares false against `nullptr`.

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,216 @@
...
+// Templated for testing/dependency injection reasons, always use the alias at the end of the file
+template <typename MutexAPI> class NamedMutexWrapper
+{
+  public:
+    NamedMutexWrapper() = delete;
+    NamedMutexWrapper(NamedMutexWrapper&) = delete;
+    NamedMutexWrapper(NamedMutexWrapper&&) = delete;
+
+    /// Some mutexes are almost never used. This makes for an easy optimization by not initializing them unless needed.
+    NamedMutexWrapper(std::wstring name, bool lazy_init = false) :
```
CarlosNihelton, line 76: Why passing the string as a local copy instead of a const-ref or a view (or even `const wchar_t*`?)

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    class Lock
+    {
+      public:
+        constexpr Lock() noexcept : parent_(nullptr), response_(0)
+        {
+        }
+
+        Lock(NamedMutexWrapper& parent) noexcept : parent_(&parent), response_(parent.wait_and_acquire())
```
CarlosNihelton, line 96: I'm a bit surprised clang-tidy didn't complain about the missing `explicit` keyword here.

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,219 @@
...
+            return parent_ && (response_ == WAIT_OBJECT_0);
+        }
+
+        DWORD why() const noexcept
+        {
+            return parent_ ? 0 : response_;
+        }
+
+        template <typename Callable> Lock&& and_then(Callable&& f)
+        {
```
CarlosNihelton, line 139: Are you sure we need to make the return types rvalue references? This seems to create complexity that may not be necessary.

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,219 @@
...
+            if (ok()) {
+                safe_execute(std::forward<Callable>(f), [&]() noexcept { release(); });
+            }
+            return std::move(*this);
+        }
+
+        template <typename Callable> Lock&& or_else(Callable&& f)
+        {
```
CarlosNihelton, line 147: Same question as above.

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,219 @@
...
+        return MutexAPI::wait_and_acquire(mutex_handle, mutex_name.c_str());
+    };
+    DWORD release() noexcept
+    {
+        return MutexAPI::release(mutex_handle, mutex_name.c_str());
+    };
+};
+
+template <typename CallableExec, typename CallablePanic>
+void safe_execute(CallableExec&& f, [[maybe_unused]] CallablePanic&& panic)
```
CarlosNihelton, line 195: Nitpick: panic is not too dramatic? Wouldn't be more appropriate to call it `onError` or `onFailure`?

CarlosNihelton (changes requested):
Very interesting piece of C++ code. Sometimes felt too advanced. Let's see.

DistroLauncher/named_mutex.cpp
```diff
@@ -0,0 +1,48 @@
...
+        const auto error = GetLastError();
+        return error != 0 ? error
+                          : ~static_cast<DWORD>(0); // Ensuring return value is always != 0 when mutex_handle is nullptr
+    }
+    return 0;
+}
+
+DWORD Win32MutexApi::destroy(HANDLE& mutex_handle, LPCWSTR mutex_name) noexcept
+{
+    if (static_cast<bool>(mutex_handle)) {
```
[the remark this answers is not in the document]
EduardGomezEscandell: Yeah, that was my idea. But then the compiler raises a warning next line because "mutex_handle might be zero". Evidently not true, but the compiler doesn't notice unless I do this.

DistroLauncher/named_mutex.h
EduardGomezEscandell: Yes, otherwise this doesn't compile:
```
mutext.lock().and_then(function).or_else(other_function);
             ^                  ^ 
Deleted constructor    Deleted constructor
```

DistroLauncher/named_mutex.h
EduardGomezEscandell: Fair enough 😄

DistroLauncher/named_mutex.cpp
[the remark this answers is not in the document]
EduardGomezEscandell: I can always use `__asume(mutex_handle != 0)`, but then clang-tidy may force me to explicitly cast it, getting back to the same problem 😓

DistroLauncher/named_mutex.cpp
EduardGomezEscandell: Good catch

DistroLauncher/named_mutex.h
EduardGomezEscandell: I though rvalue return optimization would save the day here, but apparently not, so I had to do this 🤷.

DistroLauncher/named_mutex.cpp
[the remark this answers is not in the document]
CarlosNihelton: `if(mutex_handle != NULL)` then?

DistroLauncher/named_mutex.h
EduardGomezEscandell: Yep, you also have to delete the copy-constructor:

https://godbolt.org/z/1q7EqoEWz

DistroLauncher/named_mutex.cpp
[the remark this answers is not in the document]
EduardGomezEscandell: I'll give it a try.

DistroLauncher/named_mutex.h
EduardGomezEscandell: You're right, seems like I misread the compiler at some point.

DistroLauncher/named_mutex.cpp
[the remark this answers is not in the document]
EduardGomezEscandell: Done

DistroLauncher/named_mutex.h
EduardGomezEscandell: I just found what it actually broke:
```c++
auto scope_lock = mutex.lock().and_then([] { puts("Hello!"); });
                ^ Deleted constructor Lock(Lock&)
```
Honestly I'm fine breaking this, it further discourages the scope-lock idiom. Ultimately, it can still be done with:
```c++
auto scope_lock = mutex.lock();
scope_lock.and_then([] { puts("Hello!"); });
```

DistroLauncher/named_mutex.h
```diff
@@ -0,0 +1,219 @@
...
+        {
+            return parent_ && (response_ == WAIT_OBJECT_0);
+        }
+
+        DWORD why() const noexcept
+        {
+            return parent_ ? 0 : response_;
+        }
+
+        template <typename Callable> Lock&& and_then(Callable&& f)
```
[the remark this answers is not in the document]
EduardGomezEscandell: Fixed

DistroLauncher/named_mutex.h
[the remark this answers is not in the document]
EduardGomezEscandell: Fixed

DistroLauncher/named_mutex.h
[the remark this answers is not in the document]
EduardGomezEscandell: Fixed

DistroLauncher-Tests/DistroLauncher-Tests/named_mutex_tests.cpp
```diff
@@ -0,0 +1,248 @@
...
+    return 0;
+}
+
+DWORD TestMutexApi::release(HANDLE& mutex_handle, LPCWSTR mutex_name) noexcept
+{
+    locked(mutex_handle) = false;
+    return 0;
+}
+
+std::wstring mangle_name(std::wstring name)
```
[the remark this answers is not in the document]
EduardGomezEscandell: Fixed 👍

DistroLauncher/named_mutex.cpp
EduardGomezEscandell: I re-read the documentation and this function in particular always returns NULL when failed:
> If the function succeeds, the return value is a handle to the newly created mutex object.
> If the function fails, the return value is NULL. To get extended error information, call [GetLastError](https://docs.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

THanks anyways for the heads-up because I did not know.

DistroLauncher-Tests/DistroLauncher-Tests/named_mutex_tests.cpp
```diff
@@ -0,0 +1,283 @@
...
+        {
+            TestNamedMutex mutex_2(L"test-lifetime");
+            it = std::find(dbe.cbegin(), dbe.cend(), TestNamedMutex::mangle_name(L"test-lifetime"));
+            ASSERT_NE(it, dbe.cend());  // Name still in -> destroy not called
+            ASSERT_EQ(it->refcount, 2); // create called twice
+            ASSERT_FALSE(it->locked);   // Still not locked -> wait_and_acquire not called
+        }
+
+        it = std::find(dbe.cbegin(), dbe.cend(), TestNamedMutex::mangle_name(L"test-lifetime"));
+        ASSERT_NE(it, dbe.cend());  // Name not in database -> destroy called once
```
CarlosNihelton, line 134: If I understood the test correctly the comment needs some update, because the name is still in the "database".

DistroLauncher/ProcessRunner.cpp
```diff
@@ -17,9 +17,11 @@
 
 #include "stdafx.h"
 
-namespace Helpers {
+namespace Helpers
```
CarlosNihelton, line 20: I could not grasp what is the relationship of this class with the current PR. If this is just reformattings we should deal with it in another PR.

DistroLauncher/ProcessRunner.h
```diff
@@ -22,9 +22,11 @@
 // Once created it cannot be reassigned, nor reused,
 // except for holding exit code and std out and err streams content
 // of the launched process.
-namespace Helpers {
-    class ProcessRunner {
-    private:
+namespace Helpers
```
CarlosNihelton, line 25: ditto

CarlosNihelton (changes requested):
We are almost done!

DistroLauncher-Tests/DistroLauncher-Tests/named_mutex_tests.cpp
EduardGomezEscandell: You're right. Fixed.

DistroLauncher/ProcessRunner.cpp
EduardGomezEscandell: Yeah, I was playing with pre-commit hooks earlier and didn't clean up afterwards. Reverting this.

DistroLauncher/ProcessRunner.h
EduardGomezEscandell: Samesies

PR #222, opened 2022-07-14
Better testing proposal

EduardGomezEscandell:
### Description
The idea is to link in the source files from the project, so their functionality can still be tested.

### Status
Ready

### Changelog
- Added DistroLauncher sources to Test solution
- Added onecore.lib to linker
- Added definition of `g_wslApi` in `main.cpp` (Needed by the compiler even if not used)
- Removed includes of sourcefiles
- Added a filter in VS project to keep test- and launcher-sources separate


EduardGomezEscandell:
This PR is necessary to make the `sudo.h` PR testable

PR #226, opened 2022-07-15
Patch default ubuntu-release-upgrader policy

EduardGomezEscandell:
### Dependencies
- #225, which depends on
  - #223, which depends on
      - #221 
      - #222 

### Refactor
This PR is a refactor of #213.

### Status
WIP

### Explanation

Ubuntu desktop proposes upgrades according to a certain policy:
```
$ cat /etc/update-manager/release-upgrades
# Default behavior for the release upgrader.

[DEFAULT]
# Default prompting and upgrade behavior, valid options:
#
#  never  - Never check for, or allow upgrading to, a new release.
#  normal - Check to see if a new release is available.  If more than one new
#           release is found, the release upgrader will attempt to upgrade to
#           the supported release that immediately succeeds the
#           currently-running release.
#  lts    - Check to see if a new LTS release is available.  The upgrader
#           will attempt to upgrade to the first LTS release available after
#           the currently-running one.  Note that if this option is used and
#           the currently-running release is not itself an LTS release the
#           upgrader will assume prompt was meant to be normal.
Prompt=lts
```

There are different releases of Ubuntu WSL on the Microsoft store. These can be classified in three categories. Currently, the policy is always `lts` by default. This PR changes the default upgrader policy to:

| Category   | Releases in store  | Desired default policy |
|------------|--------------------|------------------------|
| Preview    | Ubuntu (Preview)   | normal                 |
| Latest LTS | Ubuntu             | lts                    |
| Numbered   | Ubuntu 18.04.5 LTS<br>Ubuntu 20.04.4 LTS<br>Ubuntu 22.04 LTS | never |


EduardGomezEscandell:
Closing as we'll take a diferent approach

PR #229, opened 2022-07-18
Enables MSBuild to update and build the Flutter app if required

CarlosNihelton:
Attempting to build the solution without having the Flutter splash subproject checked out and compiled in release mode produces a message almost impossible to correlate with the real problem in the build system.

This PR enables automatically updating and building the Flutter submodule as part of the appx build system. The task is skipped if the `$(FlutterSplashRoot)\build\windows\runner\Release\` directory exists.


EduardGomezEscandell (approved):
Great! Always nice to make building easier.

PR #233, opened 2022-07-19
Extracts the process handling out of the splash controller

CarlosNihelton:
The splash controller is made of a state machine and a strategy class full of static functions to encapsulate the Win32 API's with the duty of starting and controlling the slide show application. There are a few premises related to its job:
1. Launcher output stream must be succefully redirected so the child process can read it and display the messages.
2. The Flutter executable must exist.
3. Starting the application must succeed.
4. Detecting its HWND also succeeds.

It's quite easy to predict that the behavior of that class could be a nightmare of if-elses. That's why the state machine design.

Yet, the strategy class that feeds the controller with the Win32 capabilities is doing too much: controlling the application process and the application window. When porting the OOBE to Windows, most of the process handling behavior would be duplicated, but the window control would not, thus splitting those duties in two different components allows for an easier migration to the new OOBE while not breaking the current behavior (required for the upcoming point release).

So, this PR extracts the process creation, termination and spurious termination notification out of the splash controller component. Also, the splash controller is updated to reflect the changes in such a way that user won't notice any difference in the application behavior.

### DIFF
--- a/DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
@@ -1,0 +1,26 @@
+#include "stdafx.h"
+#include "gtest/gtest.h"
+#include "ChildProcess.h"
+
+namespace Oobe
+{
+    const auto* fake = L"./do_not_exists";
+    TEST(ChildProcess, StartNTerminate)
+    {
+        ChildProcess process{fake, L""};
+        EXPECT_EQ(process.pid(), 0);
+        process.start();
+        EXPECT_NE(process.pid(), 0);
+        process.terminate();
+        ASSERT_EQ(process.pid(), 0);
+    }
+    TEST(ChildProcess, ListenerIsCalled)
+    {
+        bool called = false;
+        ChildProcess process{fake, L""};
+        process.setListener([&called] { called = true; });
+        EXPECT_FALSE(called);
+        process.start();
+        ASSERT_TRUE(called);
+    }
+}
--- a/DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
@@ -1,9 +1,10 @@
 #include "stdafx.h"
 #include "gtest/gtest.h"
-#include "ChildProcess.h"
+#include "FakeChildProcessImpl.h"
 
-namespace Oobe
+namespace Testing
 {
+    using ChildProcess = Oobe::ChildProcessInterface<FakeChildProcess>;
     const auto* fake = L"./do_not_exists";
     TEST(ChildProcess, StartNTerminate)
     {
--- a/DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
@@ -5,7 +5,7 @@
 namespace Testing
 {
     using ChildProcess = Oobe::ChildProcessInterface<FakeChildProcess>;
-    const auto* fake = L"./do_not_exists";
+    const auto* fake = L"./does_not_exist";
     TEST(ChildProcess, StartNTerminate)
     {
         ChildProcess process{fake, L""};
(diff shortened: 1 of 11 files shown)


DistroLauncher/ChildProcess.h
```diff
@@ -0,0 +1,164 @@
...
+            return procInfo.dwProcessId;
+        }
+
+        DWORD threadId() const
+        {
+            return procInfo.dwProcessId;
+        }
+
+        constexpr ChildProcessInterface(const std::filesystem::path& exePath, const std::wstring& args,
+                                        HANDLE stdErr = nullptr, HANDLE stdIn = nullptr, HANDLE stdOut = nullptr) :
```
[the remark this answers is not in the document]
CarlosNihelton: I'll wait for C++20 to make this nicer.

DistroLauncher/splash_controller.h
```diff
@@ -232,52 +186,19 @@ namespace Oobe
...
-                          "CallableInOtherThread callback must be void with no arguments.");
-            // Most likely the exePath will be built on the fly from a string, so there is no sense in creating a
-            // temporary and copy it when we can move.
-            ZeroMemory(&procInfo, sizeof(PROCESS_INFORMATION));
-            ZeroMemory(&startInfo, sizeof(STARTUPINFO));
-            startInfo.cb = sizeof(STARTUPINFO);
-            startInfo.hStdInput = stdIn;
-            startInfo.dwFlags |= STARTF_USESTDHANDLES;
-            SetHandleInformation(stdIn, HANDLE_FLAG_INHERIT, 1);
+            process->setListener(std::move(onClose));
```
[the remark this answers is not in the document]
CarlosNihelton: Fixed.

DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
```diff
@@ -0,0 +1,26 @@
+#include "stdafx.h"
+#include "gtest/gtest.h"
+#include "ChildProcess.h"
+
+namespace Oobe
+{
+    const auto* fake = L"./do_not_exists";
```
EduardGomezEscandell, line 7: Typo: does_not_exist

DistroLauncher-Tests/DistroLauncher-Tests/DistroLauncher-Tests.vcxproj
```diff
@@ -82,8 +82,10 @@ if EXIST "$(SharedIdb)" xcopy /Y /F "$(SharedIdb)" "$(IntDir)"</Command>
     <ClCompile Include="..\..\DistroLauncher\WindowsUserInfo.cpp" />
     <ClCompile Include="..\..\DistroLauncher\WslApiLoader.cpp" />
     <ClCompile Include="..\..\DistroLauncher\WSLInfo.cpp" />
+    <ClCompile Include="ChildProcessTests.cpp" />
     <ClCompile Include="ConsoleServiceTests.cpp" />
     <ClCompile Include="ExtendedCliParserTests.cpp" />
+    <ClCompile Include="FakeChildProcessImpl.cpp" />
```
EduardGomezEscandell, line 88: If I understand correctly, you're chosing between real and fake API by deciding what file to link? That seems like it could lead to some hard to debug issues in the future.

I would define a new class in `Testing/mock_api.h`, and define its member functions either where you did or in `Testing/mock_api.cpp`. file and put it in a namespace:
```c++
namespace testing {
    class MockWin32ChildProcess{ /* ... */ };
    using Win32ChildProcess = Win32ChildProcess<MockWin32ChildProcess>;
}

// We can distinguish by the namespace:
Oobe::Win32ChildProcess; // real one
testing::Win32ChildProcess; // mock one
```

Same as with `NamedMutex` and `Sudo`.

DistroLauncher/ChildProcess.h
```diff
@@ -0,0 +1,165 @@
...
+        {
+            return procInfo.dwProcessId;
+        }
+
+        DWORD threadId() const
+        {
+            return procInfo.dwProcessId;
+        }
+
+        constexpr ChildProcessInterface(const std::filesystem::path& exePath, const std::wstring& args,
```
EduardGomezEscandell, line 97: No `std::basic_string` constructor in c++17 is constepxr, so this function can never be called from a constexpr context.

The same is true for `std::filesystem::path`.

EduardGomezEscandell (changes requested):

EduardGomezEscandell:
Requesting changes only for the typo 😄. The others are more debatable.

DistroLauncher-Tests/DistroLauncher-Tests/ChildProcessTests.cpp
CarlosNihelton: :D

DistroLauncher-Tests/DistroLauncher-Tests/DistroLauncher-Tests.vcxproj
CarlosNihelton: You're indeed right. I took a bit more aggressive approach making the dependency more visible.

DistroLauncher/ChildProcess.h
CarlosNihelton: Indeed.

EduardGomezEscandell (approved):
I see that my "Abstract" and "Usage" comments have become a thing 😄

PR #236, opened 2022-07-21
Skip C++ source linting for unrelevant PRs

CarlosNihelton:
Linting is quite resource consuming.
Only the sources inside `DistroLauncher/` are linted.
PR's which won't touch those files (or the workflow ones) will have QA skipped.
This will not skip run-tests workflow.


EduardGomezEscandell (approved):
👍

PR #237, opened 2022-07-22
Prevents filesystem exceptions

CarlosNihelton:
Filesystem calls in the context this PR touches could fail due user making some configuration to the distro preventing Windows from accessing the referred file as well as user permissions, since the affected function is no longer called only during distro setup.

Thus, this PR avoids exceptions been thrown due filesystem API errors. `std::bad_alloc` can still be thrown :D 

I preferred to go with the `error_code` API then the Pokémon approach (`catch (...)`).


EduardGomezEscandell (approved):
Never heard of it as "Pokemon approach" but I like it 😄

PR #238, opened 2022-07-22
Adapts the image base name to the new pipeline

CarlosNihelton:
I was so happy with the new pipeline that I didn't notice the image naming scheme also changed. 😳

Older images scheme: `<codename>-server-cloudimg-<arch>-wsl.rootfs.tar.gz`.

With the new pipeline: `ubuntu-<codename>-wsl-<arch>-wsl.rootfs.tar.gz`.

Before:

image

After:

image

This PR complements #235 to adapt the image base name convention.


EduardGomezEscandell (approved):
👍

PR #239, opened 2022-07-22
Selection of default upgrade policy

EduardGomezEscandell:
This PR is a temporary solution to #226.


DistroLauncher/upgrade_policy.cpp
```diff
@@ -0,0 +1,79 @@
...
+    sed << L"sed -i " << std::quoted(regex) << L' ' << std::quoted(policyfile.wstring());
+
+    DWORD errCode;
+    auto hr = Sudo::WslLaunchInteractive(sed.str().c_str(), FALSE, &errCode);
+
+    if (SUCCEEDED(hr) && errCode == 0) {
+        std::wstringstream cmd;
+        cmd << L"date --iso-8601=seconds" << L" > " << std::quoted(log.wstring());
+        Sudo::WslLaunchInteractive(cmd.str().c_str(), FALSE, &errCode);
+    }
```
CarlosNihelton, line 72: We can use bash to execute both steps as one, including the if-logic. For instance, launching this:

```bash
bash -ec "cat does-not-exist.txt && echo Hello"
```
terminates with exit code 1 due cat filure and "Hello" is not printed.

That not only simplifies the C++ code, but mainly avoids setting and resetting the default user more than once.

DistroLauncher/upgrade_policy.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void SetDefaultUpgradePolicyImpl()
+{
+    namespace fs = std::filesystem;
+
+    const fs::path log{L"/var/log/upgrade-policy-changed.log"};
+    const fs::path policyfile{L"/etc/update-manager/release-upgrades"};
+
+    if (fs::exists(log)) {
```
CarlosNihelton, line 57: Windows will never find that file without prefixing its path with the WSL prefix.

DistroLauncher/upgrade_policy.h
```diff
@@ -0,0 +1,20 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#pragma once
+
+void SetDefaultUpgradePolicy();
```
CarlosNihelton, line 20: We need to find a way to ensure the newline at the end of the file. I personally don't care, but GitHub always complains.

CarlosNihelton (changes requested):

DistroLauncher/upgrade_policy.cpp
EduardGomezEscandell: Oops. Fixed.

DistroLauncher/upgrade_policy.cpp
```diff
@@ -54,26 +54,21 @@ void SetDefaultUpgradePolicyImpl()
...
 
-    if (fs::exists(log)) {
+    if (fs::exists(Oobe::WindowsPath(log))) {
         return;
     }
 
     std::wstring regex = L"s/Prompt=lts/Prompt=" + GetUpgradePolicy() + L'/';
     std::wstringstream sed;
     sed << L"sed -i " << std::quoted(regex) << L' ' << std::quoted(policyfile.wstring());
+    sed << L" && date --iso-8601=seconds" << L" > " << std::quoted(log.wstring());
```
CarlosNihelton, line 65: I think the command needs to be `bash -ec \' ... \' `. Did you test that this composition work in WSL? I'm afraid it would interpret the command composition as a single executable path, which would obviously fail due file not found.

DistroLauncher/upgrade_policy.cpp
EduardGomezEscandell: You're right, this is better.

DistroLauncher/upgrade_policy.h
EduardGomezEscandell: I don't know how we could do that, maybe a git hook or similar?

DistroLauncher/upgrade_policy.cpp
EduardGomezEscandell: Oh I replaced it to your sugestion before testing this, I pushed too early.

DistroLauncher/upgrade_policy.cpp
```diff
@@ -0,0 +1,80 @@
...
+        DWORD errCode;
+        Sudo::WslLaunchInteractive(command.c_str(), FALSE, &errCode);
+    }
+
+}
+
+void SetDefaultUpgradePolicy()
+{
+    auto mutex = NamedMutex(L"upgrade-policy");
+    mutex.lock().and_then(internal::SetDefaultUpgradePolicyImpl);
```
CarlosNihelton, line 79: An alternative to my previous comment about filesystem exceptions might be setting a reasonable `or_else` function here. But that seems more complicated, unless you don't want to proceed in case of filesystem errors.

CarlosNihelton (changes requested):
We are almost done, I think.

DistroLauncher/upgrade_policy.cpp
EduardGomezEscandell: Beware that `or_else` is called when we fail to switch user to root.

Throwing an exception during `and_then` will only revert the user to default and rethrow again; it will not call `or_else`.

I think the best is to have a fallback.

DistroLauncher/upgrade_policy.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    void SetDefaultUpgradePolicyImpl()
+    {
+        namespace fs = std::filesystem;
+
+        const fs::path log{L"/var/log/upgrade-policy-changed.log"};
+        const fs::path policyfile{L"/etc/update-manager/release-upgrades"};
+
+        if (fs::exists(Oobe::WindowsPath(log))) {
```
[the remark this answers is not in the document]
EduardGomezEscandell: Agree on the fallback mechanism. I used `test -f $log`.

DistroLauncher/WSLInfo.cpp
```diff
@@ -204,4 +204,32 @@ namespace Oobe
 
         return hr;
     }
+
+    std::filesystem::path WindowsPath(const std::wstring_view distro_path)
```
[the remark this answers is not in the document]
CarlosNihelton: `string_view`s are to be copied.

CarlosNihelton (approved):
LGTM!

Please do not forget to squash that one. There are a lot of less-interesting commits due merges and linting fixes overloading the commit history.

CarlosNihelton:
I merged, because I'll need one function you've implemented here.

PR #240, opened 2022-07-22
Removed Sudo::WslLaunch

EduardGomezEscandell:
This static function is problematic because the process it launches may outlive the user change.

The goal of the Sudo class is to abstract away these complexites, and this function failed to achieve this.


CarlosNihelton (approved):
LGTM

PR #242, opened 2022-07-25
Abstraction on top of Win32 Named Events that can be set only once

CarlosNihelton:
## Abstract

An abstraction on top of Win32 Named Events that can be set only once.

>NOTE: Win32 Events (named or unnamed) are resetable, meaning that one can toggle on-off as much as desired. They
can even be automatically reset by the OS, if configured to do so by the CreateEvent function. This class
restricts that behavior by closing the OS handle and nulling out the private field that would represent it, thus
rendering the object useless after being once set.

Usage
----
Ensure a unique name is used for its initialization. This is an IPC mechanism, so other processes will find it.

  ```
 SetOnceNamedEvent event(L"unique_identifyier");
 ```
 When something interesting happens and you want to let the world know about it, set the event:

 ```
bool ok = event.set();
```

That's it. From now on the variable is useless. The event remains valid from a succesfull construction until
it's set.

```
ASSERT_FALSE(event.isValid())
````

That will be used to notify the OOBE of the distro registration completion as well as to kindly ask it to quit, events that by nature happens at most once.

### DIFF
--- a/DistroLauncher/SetOnceNamedEvent.cpp
+++ b/DistroLauncher/SetOnceNamedEvent.cpp
@@ -37,7 +37,7 @@ namespace Win32Utils
         }
     }
 
-    bool SetOnceNamedEvent::set()
+    bool SetOnceNamedEvent::set() noexcept
     {
         if (event == nullptr) {
             return false;
@@ -52,10 +52,9 @@ namespace Win32Utils
 
     SetOnceNamedEvent::SetOnceNamedEvent(SetOnceNamedEvent&& other) noexcept :
         event{std::exchange(other.event, nullptr)}
-    {
-    }
+    { }
 
-    bool SetOnceNamedEvent::isValid() const
+    bool SetOnceNamedEvent::isValid() const noexcept
     {
         return event != nullptr;
     }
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: DistroLauncher/SetOnceNamedEvent.cpp [hunk 1]
```diff
@@ -0,0 +1,62 @@
...
+
+    SetOnceNamedEvent::SetOnceNamedEvent(SetOnceNamedEvent&& other) noexcept :
+        event{std::exchange(other.event, nullptr)}
+    {
+    }
+
+    bool SetOnceNamedEvent::isValid() const
+    {
+        return event != nullptr;
+    }
```
EduardGomezEscandell, line 61: noexcept

DistroLauncher/SetOnceNamedEvent.cpp
```diff
@@ -0,0 +1,62 @@
...
+        if (event == nullptr) {
+            return false;
+        }
+        if (SetEvent(event) == FALSE) {
+            return false;
+        }
+        CloseHandle(event);
+        event = nullptr;
+        return true;
+    }
```
EduardGomezEscandell, line 51: noexcept

EduardGomezEscandell (changes requested):
Not usually a purist of noexcepts, but this codebase is quite cool in that they are thoroughly used, so I'd try to keep it that way.

EduardGomezEscandell (approved):

CarlosNihelton:
The build workflow is failing due an issue in the flutter action. I'll push another PR to fix the action version temporarily until the upstream is fixed.

PR #245, opened 2022-07-25
[CI] Sets a 20 min timeout for clang-tidy

CarlosNihelton:
GitHub history shows that succesful runs finish under 15 minutes long. Longer runs required manual cancellation.
Not sure the cause, but let's limit the runs to 20 minutes for now to avoid depending on manual cancellations, at least.
GitHub default limit seems too much: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -19,6 +19,7 @@ jobs:
   quality:
     name: Lint C++ sources with LLVM
     runs-on: windows-latest
+    timeout-minutes: 20
     steps:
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell (approved):
LGTM

The longest I found in the last week or so is [this one](https://github.com/ubuntu/WSL/actions/runs/2691443185) which took 12 minutes. 20 is a good, safe choice but if we felt it necessary we could probably go down to 15 in the future.

PR #250, opened 2022-07-27
Refactor Application Strategies

CarlosNihelton:
# Why This PR

>  This refactoring won't change the current behavior.

1. To break the existing app strategies were moved to their own "components" (.h/.cpp pair).
 
As new strategies appear as new files, conditional compilation will be held by the top level header ApplicationStrategy.h, while the Application.h still presents the highest level API-only class.

2. Extract common behavior for the simpler strategies.

Those not requiring the interlocking between the installer and the slide show (aka splash screen) benefit from those implementations. When migrating the OOBE to Windows a new strategy very close to the NoSplashStrategy implemented for ARM64, but not exactly equal, would cause a lot of code duplication if those were not extracted.

## Context Introduction
From the point of view of the launcher, the OOBE consists of an `Application<>` concrete class that offers extra command line parameter handling, and `setup()`, `reconfigure()` and `runSplash()` methods.

The splash functionality only exists on x64 platforms. Currently it is an application on its own that must be started and controlled (including acquiring the knowledge of its top level window handle to allow hiding and closing that window at strategic moments). That application is fed from the launcher standard console output for a period of its execution, before any action is required from the user to be taken at the console. Thus it has a `SplashController` with its own state machine takes care of that side of the application.

The setup and reconfiguration can be done in different ways, depending on whether WSLg is available, thus an `InstallerController` represents the possibilities as a state machine, enabling the higher level application class to react to whether the TUI or GUI was choosen and close or hide the splash window at the most appropriate moment, which differs on each situation.

That complexity of how to interact with those moving parts motivated splitting those controllers and building different AppStrategies for each situation. Currently the differentiating factor between those strategies is whether running the Flutter splash window is supported or not. As we move the GUI to Windows, another orthogonal dimmension is added and more strategies will arise. The higher level Application<> class takes care of the command line composition and expects from those underlying strategies the following public interface:

```cpp
      public:
        /// Places the sequence of events to make the OOBE perform an automatic installation.
        HRESULT do_autoinstall(const std::filesystem::path& autoinstall_file);

        /// Places the sequence of events to make the OOBE perform an interactive installation.
        /// By default GUI support is checked before launching the OOBE, unless [forceTextMode]
        /// is set to true, in which case the check is skipped and the installer run in TUI mode.
        HRESULT do_install(Mode uiMode);

        /// Places the sequence of events to make the OOBE perform an automatic installation.
        HRESULT do_reconfigure();

        /// Triggers the console redirection and launch the splash screen application.
        void do_run_splash(bool hideConsole = false);
        
```


DistroLauncher/AppStrategyCommon.h
```diff
@@ -0,0 +1,37 @@
+#pragma once
```
EduardGomezEscandell, line 1: I'd stay consistent with naming and call these `ApplicationStrategyCommon.(h|cpp)` (or change others to use `App` as well).

EduardGomezEscandell (changes requested):

CarlosNihelton:
I squashed down all the changes while rebasing to avoid merge conflicts. Essentially, the last changes were the renamings you suggested and `using` instead of `#define`.

DistroLauncher-Tests/DistroLauncher-Tests/DistroLauncher-Tests.vcxproj
```diff
@@ -58,14 +58,15 @@
...
-      <Command>if EXIST "$(SharedPdb)" xcopy /Y /F "$(SharedPdb)" "$(IntDir)"
+      <Command>if EXIST "$(SharedPdb)" xcopy /Y /F "$(SharedPdb)" "$(IntDir)"
 if EXIST "$(SharedIdb)" xcopy /Y /F "$(SharedIdb)" "$(IntDir)"</Command>
       <Outputs>$(IntDir)vc$(PlatformToolsetVersion).pdb;</Outputs>
       <Inputs>$(SharedPdb)</Inputs>
     </CustomBuildStep>
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="..\..\DistroLauncher\ApplicationStrategy.cpp" />
+    <ClCompile Include="..\..\DistroLauncher\AppStrategyCommon.cpp" />
```
EduardGomezEscandell, line 69: Test is breaking here I think

DistroLauncher-Tests/DistroLauncher-Tests/DistroLauncher-Tests.vcxproj.filters
```diff
@@ -101,6 +98,7 @@
     </ClCompile>
     <ClCompile Include="..\..\DistroLauncher\LauncherForceMode.cpp" />
     <ClCompile Include="LauncherForceModeTests.cpp" />
+    <ClCompile Include="..\..\DistroLauncher\AppStrategyCommon.cpp" />
```
EduardGomezEscandell, line 101: Also here

CarlosNihelton: That's right.

EduardGomezEscandell (approved):

PR #251, opened 2022-07-27
Models the environment variable LAUNCHER_FORCE_MODE

CarlosNihelton:
That variable is used to force a UI mode from the outside, before running the launcher. It's meant mainly for testability.

```cpp
    ///	0 or unset or invalid = autodetection
    ///	1 = text mode
    ///	2 = GUI mode.
```

It's use was located in the implementation of the function `mustRunOOBEinTextMode()` tightly coupled with detecting WSLg support. That function was in turn called by the `installer_policy.h` and the already dead `OOBESetup()` function.

The OOBE on Windows will have other requirements, but not WSLg, thus decoupling is necessary in order to preserve the env var semantics.

Thus, let's promote it to a "component" (its own thing) and decouple it from WSLg detection. The InstallerPolicy knows what to do if the variable is set or not, the same will be for the Windows OOBE application strategy.

This also clears out the consequently dead code plus a TODO around naming: moving the function `WslGraphicsSupported` to the `Oobe::internal` namespace, instead of  the `Helpers` one, which belongs to the upstream source code and functionality. That function usage is now limited to one single caller.


EduardGomezEscandell (approved):
I like decoupling, and I like extra testing 👍

PR #253, opened 2022-07-28
Let SetOnceNamedEvent wait for confirmation

CarlosNihelton:
Calling SetEvent may return quickly enough so that the event HANDLE is closed without the other end receiving the notification.
SignalObjectAndWait let's us wait for the same event we are setting, so that we only proceed when the OS really delivered the notification.


EduardGomezEscandell (approved):
👌

PR #254, opened 2022-07-28
New app strategy for ARM64 in the context of the new OOBE

CarlosNihelton:
It is expected that UDI snap will be replaced by Subiquity only.
Thus GUI will only be supported if Flutter is supported on Windows.
Thus, for ARM64, only TUI will be possible.
Although we could add some checks to the existing NoSplashStrategy,
I'd rather keep them orthogonal as other differences may be detected later on.

The next step is the x64 strategy, which is more complicated, thus I preferred to push on a separate PR.

### DIFF
--- a/DistroLauncher/WinTuiStrategy.cpp
+++ b/DistroLauncher/WinTuiStrategy.cpp
@@ -9,9 +9,9 @@ namespace Oobe
         return internal::do_autoinstall(installer, autoinstall_file);
     }
 
-    HRESULT WinTuiStrategy::do_install(Mode ui)
+    HRESULT WinTuiStrategy::do_install(Mode uiMode)
     {
-        if (ui == Mode::Gui) {
+        if (uiMode == Mode::Gui) {
             wprintf(L"GUI mode is not supported on this platform.\n");
         }
         return internal::install_linux_ui(installer, Mode::Text);
--- a/DistroLauncher/WinTuiStrategy.cpp
+++ b/DistroLauncher/WinTuiStrategy.cpp
@@ -1,0 +1,29 @@
+#include "stdafx.h"
+#include "ApplicationStrategyCommon.h"
+#include "WinTuiStrategy.h"
+
+namespace Oobe
+{
+    HRESULT WinTuiStrategy::do_autoinstall(const std::filesystem::path& autoinstall_file)
+    {
+        return internal::do_autoinstall(installer, autoinstall_file);
+    }
+
+    HRESULT WinTuiStrategy::do_install(Mode ui)
+    {
+        if (ui == Mode::Gui) {
+            wprintf(L"GUI mode is not supported on this platform.\n");
+        }
+        return internal::install_linux_ui(installer, Mode::Text);
+    }
+
+    HRESULT WinTuiStrategy::do_reconfigure()
+    {
+        return internal::reconfigure_linux_ui(installer);
+    }
+
+    void WinTuiStrategy::do_run_splash(bool hideConsole)
+    {
+        wprintf(L"This device architecture doesn't support running the splash screen.\n");
+    }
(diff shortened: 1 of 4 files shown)


DistroLauncher/WinTuiStrategy.h
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
EduardGomezEscandell, line 17:
```suggestion
/*
 * Copyright (C) 2022 Canonical Ltd
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License version 3 as
 * published by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 *
 */
 
 #pragma once
```
Flipping order for consistency.

DistroLauncher/WinTuiStrategy.cpp
```diff
@@ -0,0 +1,29 @@
+#include "stdafx.h"
```
EduardGomezEscandell, line 1:
```suggestion
/*
 * Copyright (C) 2022 Canonical Ltd
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License version 3 as
 * published by the Free Software Foundation.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 *
 */
 
#include "stdafx.h"
```
Missing copyright header. I don't actually know how necessary this is, just going by imitation.

DistroLauncher/WinTuiStrategy.cpp
```diff
@@ -0,0 +1,29 @@
...
+{
+    HRESULT WinTuiStrategy::do_autoinstall(const std::filesystem::path& autoinstall_file)
+    {
+        return internal::do_autoinstall(installer, autoinstall_file);
+    }
+
+    HRESULT WinTuiStrategy::do_install(Mode uiMode)
+    {
+        if (uiMode == Mode::Gui) {
+            wprintf(L"GUI mode is not supported on this platform.\n");
```
EduardGomezEscandell, line 15: (Non blocking)
I know this uses `wprintf` for consistency reasons, but do we want to change this in the future? Just some ideasd.
1. Sending to stderr
2. Useing iostream
3. Implementing or importing an existing logger at some point

EduardGomezEscandell (changes requested):
Teeny tiny nitpicks

DistroLauncher/WinTuiStrategy.cpp
CarlosNihelton: I've been dreaming with a logging facility. When the time comes, it will essentially be ` s/wprintf(\(.+\));/LOG(\1);/g
` .

PR #255, opened 2022-07-29
Implements the WinOobeStrategy

CarlosNihelton:
## Abstract

Complementary to #254 , this PR implements the component will feed the Application class with the behavior necessary to run the OOBE on Windows. No dependency on WSLg. Requires x64 platform.

Choosing the app strategy now depends on:

- the platform set by the compiler
- the definition of WIN_OOBE_ENABLED constant (not yet defined anywhere, thus this won't affect 22.04.1 release).

Since the top-level `Application<>` class remains the same both the supported command line and its semantics must be preserved.

### TUI

Terminal user interface support is still required because there are still situations only detectable during runtime that may prevent running the new OOBE, such as user disabling WSL2 localhost forwarding in `~\.wslconfig`. Achieving TUI support is very similar to what is done in `SplashEnabledStrategy`, with the need to present the splash and close it when the TUI is ready for interaction. Thus, the `InstallerController` component is reused as is, even though it can only be used in TUI mode. In the near future we will be able to further reduce its functionality when we completely drop support for running the OOBE thrrough WSLg, but that's topic for another cycle, when we would be sure it won't be anymore necessary.

#### Command line
As with SplashEnabledStrategy, this component must offer the slide show (aka splash) and control the installer. As the OOBE on Windows is merged with the slide show as a single app, the installer is started at the `do_run_splash()` method. That means that the required command line must be known at that moment. `makeCli()` thus enables creating the command line for the OOBE by ensuring a minimal stem concatenated with further options passed as parameters. The command line for the setup may have an optional `--prefil` option, if the launcher succesfully acquire some user information necessary to prefill the UI fields during setup (user name and language). In such scenario we fill a file buffer with that information before starting the OOBE and write its contents into the rootfs after registration.

#### Setup versus Reconfiguration
`do_run_splash()` only runs for the WSL setup variant. The same application is also used for the reconfiguration variant. So, `do_reconfigure()` must be able to start the OOBE process as well, with different command line options. `do_run_splash()` and `do_reconfigure()` are never called in the same run of the launcher. In that situation, there is no need for the slide show, as the distro is already registered, but that's handled in the Flutter code.

The runtime requirements for both variants are the same but their behaviors are very different, so the function `run_on_autodetect_ui()` provides selecting which UI to run and accept callables with the different behaviors that must be achieved.

Setup UI can be selected via launcher command line (via the option `--ui`) as well as by setting the `LAUNCHER_FORCE_MODE` environment variable (reading it is part of the autodetection routine). For simplicity it was decided in the past that we wouldn't support the `--ui` option in reconfiguration, but instead rely on autodetection. That remains.

#### Server control and monitoring

In GUI mode everything server-related is done by Flutter. This is a great point of simplification, since we don't need to start it, spawn small processes to watch its startup and timeout after some attempts. Timeout heuristics are always questionable.

#### GUI Startup and IPC

The slide show reads text from standard input that comes from the launcher's stdout to monitor for errors, as before. Launcher is responsible for piping stdout to the child process stdin to make that possible. The same `ConsoleService` used in`SplashEnabledStrategy` fed by a named pipe is used for that.

Additionally, the launcher must notify the child process of the suitable moment to start the server (after registration is complete) and request it to gracefully quit, if TUI is required instead of GUI, for instance. Both behaviors are achieved by using distinct `SetOnceNamedEvent`s created before running the OOBE and set when (and if) necessary. Closing the OOBE window is achieved by setting one named event (support for that is landing in https://github.com/canonical/ubuntu-desktop-installer/pull/1032).

## Trying out

Playing with this requires cloning https://github.com/canonical/ubuntu-desktop-installer somewhere, compiling the Windows entry point of the ubuntu_wsl_setup package, copying the generated artifacts into this repository build folder (`x64\Debug` for instance) and passing the `WIN_OOEB_ENABLED` definition during compilation. Quite complicated for now, but my next PRs will address that complexity by bringing UDI as a submodule and tweaking the build system to select and build the correct submodule automatically as well as defining the constant.

```
cd <this-repository-root-dir>
git clone https://github.com/canonical/ubuntu-desktop-installer UDI
cd UDI/packages/ubuntu_wsl_setup
flutter pub get
flutter build windows -t lib/main_win.dart --release
copy build/windows/runner/Release/* <this-repository-root-dir>/x64/Debug
cd <this-repository-root-dir>
.\DistroLauncher.sln
```
Set WIN_OOBE_ENABLED in the IDE GUI and debug it as you would.

### DIFF
--- a/DistroLauncher/WinOobeStrategy.cpp
+++ b/DistroLauncher/WinOobeStrategy.cpp
@@ -15,7 +15,7 @@ namespace Oobe
         return Win32Utils::homedir() / L".wslconfig";
     }
 
-    std::wstring createEventName(const wchar_t const* suffix)
+    std::wstring createEventName(const wchar_t* const suffix)
     {
         return L"Local\\" + DistributionInfo::Name + L'-' + suffix;
     }
@@ -312,6 +312,7 @@ namespace Oobe
     WinOobeStrategy::WinOobeStrategy() :
         oobeExePath{getOobeExePath()}, hRegistrationEvent{createEventName(L"registered").c_str()},
         hCloseOobeEvent{createEventName(L"close-oobe").c_str()}, prefill{}
-    { }
+    {
+    }
 
 }
--- a/DistroLauncher/WinOobeStrategy.cpp
+++ b/DistroLauncher/WinOobeStrategy.cpp
@@ -124,8 +124,8 @@ namespace Oobe
         if (!oobeProcess.has_value()) {
             return E_UNEXPECTED;
         }
-        auto& oobe = oobeProcess.value();
-        if (!oobe.start()) {
+
+        if (!oobeProcess->start()) {
             return E_APPLICATION_ACTIVATION_EXEC_FAILURE;
         }
 
@@ -133,7 +133,7 @@ namespace Oobe
             do_close_oobe();
             return EVENT_E_USER_EXCEPTION;
         }
-        if (oobe.waitExitSync() != 0) {
+        if (oobeProcess->waitExitSync() != 0) {
             return E_FAIL;
         }
         return S_OK;
@@ -272,7 +272,6 @@ namespace Oobe
             wprintf(L"Failed to prepare the console configuration.\n");
             return;
         }
-        auto& console = consoleService.value();
 
         std::unique_lock<std::timed_mutex> guard{consoleGuard, std::defer_lock};
         using namespace std::chrono_literals;
@@ -282,7 +281,7 @@ namespace Oobe
             return;
         }
 
-        console.redirectConsole();
+        consoleService->redirectConsole();
 
         prefill = WslFileBuf{DistributionInfo::GetPrefillInfoInYaml(), L"/var/log/prefill-system-setup.yaml"};
         const auto cli{prefill.isEmpty() ? makeCli() : makeCli(L" --prefill=", prefill.linuxPath)};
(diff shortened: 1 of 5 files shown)


DistroLauncher/WinOobeStrategy.cpp
```diff
@@ -0,0 +1,317 @@
...
+    {
+        return Win32Utils::thisAppRootdir() / L"ubuntu_wsl_setup.exe";
+    }
+
+    std::filesystem::path getWslConfigPath()
+    {
+        return Win32Utils::homedir() / L".wslconfig";
+    }
+
+    std::wstring createEventName(const wchar_t const* suffix)
```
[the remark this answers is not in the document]
CarlosNihelton: Placed the const at the wrong location :D

DistroLauncher/WinOobeStrategy.cpp
```diff
@@ -0,0 +1,318 @@
...
+    // since the requirements for running the OOBE won't change between them.
+    template <typename OnGui, typename OnTui, typename OnNone>
+    HRESULT run_on_autodetect_ui(OnGui onGui, OnTui onTui, OnNone onNone)
+    {
+        static_assert(std::is_convertible_v<OnGui, std::function<HRESULT()>>,
+                      "onGui callback must be a callable returning HRESULT with no arguments");
+        static_assert(std::is_convertible_v<OnTui, std::function<HRESULT()>>,
+                      "onTui callback must be a callable returning HRESULT with no arguments");
+        static_assert(std::is_convertible_v<OnNone, std::function<HRESULT()>>,
+                      "onNone callback must be a callable returning HRESULT with no arguments");
```
EduardGomezEscandell, line 55: C++20's concepts would be nice here. Some day 😄.

DistroLauncher/WinOobeStrategy.cpp
```diff
@@ -0,0 +1,318 @@
...
+    {
+        // The OOBE process must not be previously created.
+        assert(!oobeProcess.has_value());
+        const auto cli = makeCli(L"--reconfigure");
+        oobeProcess.emplace(oobeExePath, cli.c_str(), nullptr, nullptr, nullptr);
+        if (!oobeProcess.has_value()) {
+            return E_UNEXPECTED;
+        }
+        auto& oobe = oobeProcess.value();
+        if (!oobe.start()) {
```
EduardGomezEscandell, line 128: You can use `obeeProcess->start()` if you're doing this to avoid calling `.value()` every time.

DistroLauncher/WinOobeStrategy.cpp
CarlosNihelton: I'd love to jump straight into C++23 to use `std::unreachable`, `std::expected` and more `constexpr`, besides C++20's `std::source_location`, `std::format`, and my favourite: C19 struct designated initialization. C++ source code will look very different after those.

DistroLauncher/WinOobeStrategy.cpp
CarlosNihelton: Good point. I think it's clearer, even though it looks like a pointer.

CarlosNihelton:
Rebasing due merge conflicts.

PR #258, opened 2022-07-29
Traverse the filesystem only once when looking for UDI and Subiquity snaps.

CarlosNihelton:
Here is the key of this PR:

```diff
        // if none of the required snaps exist in the rootfs no UI will be available.
-        if (!internal::hasUdiSnap() && !internal::hasSubiquitySnap()) {
+        if (!internal::hasAnyOfSnaps(L"ubuntu-desktop-installer"sv, L"subiquity"sv)) {
```

For that to be possible, `hasAnyOfSnaps()` has to be implemented as a template function that can accept a variadic number of `string_view` arguments and applies the existing algorithms to find entries inside the snap directory matching one of the supplied arguments. Since that receives a variadic argument in order to nicely compose `find_file_if` and `std::any_of` we put together a `push_back_many` algorithm that populates a vector of string_views from a variadic argument list.

BTW, `find_file_if` returns a bool instead of iterator, so more like `std::any_of` than `std::find_if`, thus the function was renamed to `any_file_of`.

And since there were some documentation comments lacking on the `algorithms.h` header and some unevitable linter complains about parameters of the same type in `starts/ends_with` algorithms, I put some comments to stop linter at those lines.

I think it is easier to review this PR commit by commit.

### DIFF
--- a/DistroLauncher/WSLInfo.cpp
+++ b/DistroLauncher/WSLInfo.cpp
@@ -141,7 +141,7 @@ namespace Oobe::internal
     {
         namespace fs = std::filesystem;
         const auto path = Oobe::WindowsPath(fs::path{L"/var/lib/snapd/snaps/"});
-        return find_file_if(path, [name](const auto& entry) {
+        return any_file_of(path, [name](const auto& entry) {
             return fs::is_regular_file(entry) && starts_with({entry.path().filename().wstring()}, name);
         });
     }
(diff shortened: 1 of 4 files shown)


EduardGomezEscandell (changes requested):
Minor performance comment

EduardGomezEscandell (approved):

PR #259, opened 2022-08-01
Refactored starts_with and ends_with

EduardGomezEscandell:
## Abstract
Removed the need to add braces around string literals in `starts_with` and `ends_with`:
```diff
- starts_with(string, { L"string_literal" });
+ starts_with(string,   L"string_literal"  );
```

## Explanation
The main issue is that `std::end(const char*)` points to the entry after the null termination, making naive implementations fail. This can be solved using string_views. Then template parameter deduction starts behaving funny. This is solved here with a helper adaptor template that does the following type conversions, in this order:
- `const CharT* -> basic_string_view<CharT>`: SFINAE'd to work only with chars. Null-termination expected.
- `const T[N] -> basic_string_view<T>`: SFINAE'd to work for non-characters only. No null-termination.
- `const T& -> const T&`: Any other containers stay as they were.

Comparing `char` and `wchar_t` (or any other combination of diferent chars) stays forbidden, now with a static assert instead of template resolution falure.

### Side benefits
These were not the goal but it's nice to have them
- Expanded functionality: These two functions now work with any container (See unit tests with arrays, vectors and c-style arrays of int).
- Deduplication: `starts_with` and `ends_with` now use the same implementation `starts_with_impl`, simply switching between forward and reverse iterators.

### DIFF
--- a/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
@@ -58,6 +58,27 @@ TEST(AlgorithmTests, StartsEndsWithExtra)
 
     ASSERT_FALSE(starts_with(vec_1, arr_2));
     ASSERT_TRUE(ends_with(vec_1, arr_2));
+
+    int carray[] = {1, 1, 2, 3, 5, 8, 13};
+    ASSERT_TRUE(starts_with(carray, arr_1));
+    ASSERT_FALSE(ends_with(carray, arr_1));
+
+    ASSERT_FALSE(starts_with(carray, arr_2));
+    ASSERT_TRUE(ends_with(carray, arr_2));
+
+    int* cvector = carray;
+    ASSERT_TRUE(starts_with(cvector, arr_1));
+    ASSERT_FALSE(ends_with(cvector, arr_1));
+
+    ASSERT_FALSE(starts_with(cvector, arr_2));
+    ASSERT_TRUE(ends_with(cvector, arr_2));
+
+    const int* c_const_vector = carray;
+    ASSERT_TRUE(starts_with(c_const_vector, arr_1));
+    ASSERT_FALSE(ends_with(c_const_vector, arr_1));
+
+    ASSERT_FALSE(starts_with(c_const_vector, arr_2));
+    ASSERT_TRUE(ends_with(c_const_vector, arr_2));
 }
 
 TEST(AlgorithmTests, Concat)
--- a/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
@@ -65,8 +65,6 @@ TEST(AlgorithmTests, StartsEndsWithExtra)
 
     ASSERT_FALSE(starts_with(carray, arr_2));
     ASSERT_TRUE(ends_with(carray, arr_2));
-
-    starts_with(L"Hi! How are you?", "Hi!");
 }
 
 TEST(AlgorithmTests, Concat)
--- a/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
@@ -65,6 +65,12 @@ TEST(AlgorithmTests, StartsEndsWithExtra)
 
     ASSERT_FALSE(starts_with(carray, arr_2));
     ASSERT_TRUE(ends_with(carray, arr_2));
+
+    char str_1[] = "Is";
+    ASSERT_TRUE(starts_with("Is this the real life?", str_1));
+    
+    char str_2[] = "Is this just fantasy?";
+    ASSERT_TRUE(ends_with(str_2, "fantasy?"));
 }
 
 TEST(AlgorithmTests, Concat)
--- a/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
+++ b/DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
@@ -17,8 +17,9 @@
 
 #include "stdafx.h"
 #include "gtest/gtest.h"
+#include "mock_api.h"
 
-TEST(UpgradePolicyTests, StringParsing)
+TEST(AlgorithmTests, StartsWith)
 {
     ASSERT_FALSE(starts_with(L"", L"hello"));
     ASSERT_FALSE(starts_with(L"he", L"hello"));
@@ -26,7 +27,14 @@ TEST(UpgradePolicyTests, StringParsing)
     ASSERT_TRUE(starts_with(L"hello, world!", L"hello"));
     ASSERT_FALSE(starts_with(L"cheers, world!", L"hello"));
     ASSERT_FALSE(starts_with(L"HELLO", L"hello"));
+    ASSERT_FALSE(starts_with(L"", L"Ubuntu"));
 
+    const std::wstring test_str{L"Ubuntu 22.04.1 LTS"};
+    ASSERT_TRUE(starts_with(test_str, L"Ubuntu"));
+}
+
+TEST(AlgorithmTests, EndsWith)
+{
     ASSERT_FALSE(ends_with(L"", L"world!"));
     ASSERT_FALSE(ends_with(L"d!", L"world!"));
     ASSERT_TRUE(ends_with(L"world!", L"world!"));
@@ -35,12 +43,24 @@ TEST(UpgradePolicyTests, StringParsing)
     ASSERT_FALSE(ends_with(L"this string is completely diferent", L"world!"));
     ASSERT_FALSE(ends_with(L"HELLO", L"hello"));
 
-    ASSERT_FALSE(starts_with(L"", L"Ubuntu"));
-    ASSERT_TRUE(starts_with(L"Ubuntu 22.04.1 LTS", L"Ubuntu"));
-    ASSERT_TRUE(ends_with(L"Ubuntu 22.04.1 LTS", L"LTS"));
+    const std::wstring test_str{L"Ubuntu 22.04.1 LTS"};
+    ASSERT_TRUE(ends_with(test_str, L"LTS"));
 }
 
-TEST(UpgradePolicyTests, Concat)
+TEST(AlgorithmTests, StartsEndsWithExtra)
+{
+    const std::vector<int> vec_1{1, 1, 2, 3, 5, 8, 13};
+    const std::array<int, 3> arr_1{1, 1, 2};
+    const std::array<int, 3> arr_2{5, 8, 13};
+
+    ASSERT_TRUE(starts_with(vec_1, arr_1));
+    ASSERT_FALSE(ends_with(vec_1, arr_1));
+
+    ASSERT_FALSE(starts_with(vec_1, arr_2));
+    ASSERT_TRUE(ends_with(vec_1, arr_2));
+}
+
+TEST(AlgorithmTests, Concat)
 {
     // Checking default functionality
     std::wstring dog = L"dog";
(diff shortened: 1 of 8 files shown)


DistroLauncher/algorithms.h
```diff
@@ -18,40 +18,91 @@
...
+
+    template <typename Char>
+    constexpr std::enable_if_t<is_character<Char>(), std::basic_string_view<const Char>> string_adaptor(
+      Char const* value)
+    {
+        return {value};
+    }
+
+    template <typename IteratorTested, typename IteratorCompare>
+    bool starts_with_impl(IteratorTested tested_begin, IteratorTested tested_end, IteratorCompare compare_begin,
```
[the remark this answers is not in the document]
EduardGomezEscandell: clang-format must hate the standard library then 😓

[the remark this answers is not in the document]
CarlosNihelton: I don't disagree that those parameters are easily swappable and that would cause a silent bug. That's why named parameter support in C++ would be handy. And that's why C99/C++20 struct designated initializers are handy, because they allow mimicking named parameters. Even nicer it would be if the order didn't match. I hope C++ adopts such soon.

DistroLauncher/WSLInfo.h
```diff
@@ -72,7 +72,7 @@ namespace Oobe::internal
             }
             const std::wstring filename{entry.path().filename().wstring()};
             return std::any_of(std::begin(elements), std::end(elements), [&filename](const auto& element) {
-                return starts_with({filename}, element) && ends_with({filename}, L".snap"sv);
+                return starts_with(filename, element) && ends_with(filename, L".snap"sv);
```
CarlosNihelton, line 75: I think the `L".snap"sv` can drop the literal suffix, can't it?

EduardGomezEscandell: Yes indeed. Also the `using namespace std::literals::string_view_literals
` directive. Changed.

DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
```diff
@@ -35,12 +43,37 @@ TEST(UpgradePolicyTests, StringParsing)
     ASSERT_FALSE(ends_with(L"this string is completely diferent", L"world!"));
     ASSERT_FALSE(ends_with(L"HELLO", L"hello"));
 
-    ASSERT_FALSE(starts_with(L"", L"Ubuntu"));
-    ASSERT_TRUE(starts_with(L"Ubuntu 22.04.1 LTS", L"Ubuntu"));
-    ASSERT_TRUE(ends_with(L"Ubuntu 22.04.1 LTS", L"LTS"));
+    const std::wstring test_str{L"Ubuntu 22.04.1 LTS"};
+    ASSERT_TRUE(ends_with(test_str, L"LTS"));
```
CarlosNihelton, line 47: Supporting heterogeneous string-like entities is a great thing! :D

DistroLauncher-Tests/DistroLauncher-Tests/algorithms_tests.cpp
```diff
@@ -63,4 +96,4 @@ TEST(UpgradePolicyTests, Concat)
     std::filesystem::path example_file{L"/home/fox/documents/example.json"};
     auto with_path = concat(L"diff ", example_file, L" ", example_file.wstring()); // Only first one to be quoted
     ASSERT_EQ(with_path, LR"(diff "/home/fox/documents/example.json" /home/fox/documents/example.json)");
-}
+}
```
CarlosNihelton, line 99: That remove-newline-at-eof-thingy is really annoying. We'll soon need to implement a way to avoid it.

DistroLauncher/algorithms.h
```diff
@@ -18,40 +18,80 @@
...
+namespace algo_internal
 {
-    if (tested.size() < start.size()) {
-        return false;
+    // Tells chars appart from other types
+    template <typename T> constexpr bool is_character() noexcept
+    {
+        return std::is_same_v<T, char> || std::is_same_v<T, wchar_t> || std::is_same_v<T, signed char> ||
+               std::is_same_v<T, unsigned char> || std::is_same_v<T, char16_t> || std::is_same_v<T, char32_t>;
+        // C++20 introduces char8_t
```
CarlosNihelton, line 28: Did you consider checking if there is a specialization of `std::char_traits<T>::char_type`  to avoid such chain of `is_same_v` and get it automatically up-to-date with the standard library version in use?

CarlosNihelton: We could use something like:

```cpp
template <typename CharT>
constexpr bool is_character() {
    // We know for sure double won't have a char_traits specialization.
    using default_type = std::char_traits<double>::int_type;
    return !std::is_same_v<std::char_traits<CharT>::int_type, default_type>;
}
```

[See further details on godbolt.org](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIM6SuADJ4DJgAcj4ARpjE/lykAA6oCoRODB7evv5JKWkCIWGRLDFxZgl2mA7pQgRMxASZPn4BldUCtfUEhRHRsfG2dQ1N2a1D3aG9Jf3lAJS2qF7EyOwc5gDMocjeWADUJutuCgTEocAH2CYaAIIbWzuY%2B4dOx8SYrBdXt2abDNteewObicLEMeESnxuXwImBYiQMMKebgIAE9EoxWI83Ah6gAVSHXNAMY6YVSJYi7KKoTy7PAKAD6yBxxCYDliEFm%2BwA7FYbrt%2BbsAPSC3YAdUeAGsGKgAO67fgUhRLR7oRZRei7GUCMAcAi7HEAN0eTF2TPq9JOTEICl2CnRyDwYjwAC8mI4BAA6L4C3ZeVJGXZYKhMLy0AgWtGPA4AEVtBHQIBQzItLOtQNVXnVmAuidC4dR6IOvOuPuOCZQiz1QKB%2BzMZhMAFZkQBJQS7AuPBVI8xmbuHduRvDoCBlxNm4gpq0EBRA7F4nNJ80d2Ye5hsDlF70C0cVrxVw41nsXWk2jvdut9twD9FDkfxsfJy1pw5z4j49bYXOCCPoldrzAcpetb1h%2BjZuAwR7rMWPpvAQSwMLsYBgDudL0gomL0gaQI7uOk7Pm4r7vp%2BIB5j%2BmCkIGmDBqG%2BaRp8UFfCYXLRoxNx5rsoKhIBTHQdu967vubiHnWYEQU2YnIriAAS2AAErYGKCkAOIAPI9kBO5Up4YiJDim58nx5ZoHuQFHuJr4gMBQGoeOrIwsQQLjhc3EHv2ZngdGariFZ1b9jZzJ2bE6ZedmH4bgxBn8jhlamSJ4mtgQlnqb5V7%2BfUgUOc8gjOZyKXAaJchBAIoA%2Ba5qUMrZbKZW4XjEngwBhOguy0MVOX6SWhljjFeXuQwopDlizJJReeVpSyVVAjKuEEDlsX1uJABiLVusNvajRVAUTYcVDLTNYWzO1pb8cZgnCfN4GvrsXAAGyrdZG3pVtbjjjdFqzT1cUXcyuzrGYd3rYym32Y5zK/W9%2B2HZ1AlzaJKGvO8LD/WVJ6A49wOHHDJwI%2B9yO9QtzZBNgABUSNCX5D3jejbj44TRNtRFtzMRw8y0JwDa8H4HBaKQqCcEJljWLaizLFGPw8KQiVc8z8wSiA11mB6N0AJzXesXLXVwv1Kw2kgNvonCSBzmi8LzHC8AoIAaBLxvzHAsAwIgFZwnQsTkJQaDO/QcQGsgiSJJhXBK/SCKYMc9KqNd0g0GGsQWxAUTG6QUShPUKKcOLHtsIIKkMLQadS6QWCgkY4gF/gbzVEaFsF6SVR7qs3N5pgrPcLwtB4FELLECiHhYInJx4IjrfzDtTDAAoABqeCYDKKnopz4v8IIIhiOwUgyIIigqOoBe6AkBhGCg1jWPoHcW5A8yoIk7rEpwgoqU1gqjqYAuWFwXLm83VQ3y4DDuJ4zQ9DBEmMUUoehkipBviMPwCQIH5AYD0UB/QKhf3aAwTowwAHZBQfYG%2BGCJhFD6HECo4xoF6GOF0RBRCJDzAUMLFYegsarB4CzNmRsC6mwjpIDiCgfa7ANIHD0Idji7AgPzKwlgKK4EICQWs6wEi7A8J7WIciuCzF4JLLQswZYgAbFbFuhtSCI0kOsD0AAODQSsuBcDMddBsZiuQNi5JIa6pBObc1NubS21spa2wdhAJAlZEh7jdhAD2iQXbEHCJiTgXCeF8IEUrIRbpQ4EF4JgfARBTgJgSEvYQohxDrzyVvNQic96kBlCyRI6dmb6w4OzNxidTYqT3MEvUqAqC7DiSwXhyB%2BGCOEXqMRqBlEUg2GojRNt5gIHeFgOIHI6mGMRg2a6islYaDMWYzWkguBmEkGYHWjSOGcC8VbTR0tSCyzVuY9YGh1jy3WLYrkGhdZ6xbusdhHiTk%2BK0awjgZhPkm2%2Bec7RpAjTEFSM4SQQA%3D%3D)

DistroLauncher/algorithms.h
```diff
@@ -18,40 +18,80 @@
...
+     * - T[N] into basic_string_view<T> (works for non-chars too!)
+     * - CharT* into basic_string_view<CharT>
+     * - Anything else stays the same
+     */
+    template <typename T> constexpr T const& view_adaptor(T const& value)
+    {
+        return value;
+    }
+
+    template <std::size_t Size, typename T> constexpr auto view_adaptor(const T value[Size])
```
CarlosNihelton, line 42: The array will always decay into a pointer, thus this case is never instantiated, which seems to be actually fine, otherwise this would be a bug:

```cpp
const std::string tested{"Ubuntu 22.04 LTS"};
const char sub[7] = "Ubuntu"; //6 chars + \0.
// would fail because `std::basic_string_view<const char>(sub, 7);` would forcebly include the '\0' in the view, then `mismatch.first` would be an iterator to the last char, not the one-after-last.
ASSERT_TRUE(starts_with(tested, sub));
```
If you comment out the pointer specialization you'll see errors due the '\0' being included in the string_view due the array specialization.

I think this must be removed.

CarlosNihelton (changes requested):
One minor technical issue.

DistroLauncher/algorithms.h
EduardGomezEscandell: This is on purpose. `char`-like types decay to a pointer which is what we want, but non-chars don't see the declaration of the pointer version (due to the `enable_if`) and hence stay as c-arrays. I can explicitly disable it for `char`-like types, though.

DistroLauncher/algorithms.h
EduardGomezEscandell: I played around for a while. Signed and unsined characters fail the test:

https://godbolt.org/z/M17csTEc1

I also tried using an empty struct instead of double but it also has `long` as its integer type. I understand that is the default value.

The concatenated `is_same_v`s is not pretty but it has going for it that it's very transparent as to how it works.

DistroLauncher/algorithms.h
EduardGomezEscandell: Done, now pointer decay is only enabled for chars; and c-array is only enabled for non-chars:
```c++
template<typename T>
enable_if_t<!is_character<T>(), basic_string_view<T>> view_adaptor(const T value[Size]);

template<typename CharT>
enable_if_t<is_character<CharT>(), basic_string_view<CharT>> view_adaptor(const CharT* value)
```

The need for concepts strikes once again 😓

DistroLauncher/algorithms.h
CarlosNihelton: I like it better than the previous alternative!

CarlosNihelton (approved):
Thanks for taking that work!

PR #260, opened 2022-08-02
Upload program debug databases

CarlosNihelton:
On top of the current practices of:
 1. uploading some build artifacts during `build-wsl` CI workflow and
 2. saving some metadata inside the repository wiki for further reuse during other runs of that workflow.

The proposed set of changes allows collecting program database files (.PDB) generated during compilation of the binaries bundled in the appx for easier crash analysis in the future.

PDBs are collected for each app built, always, and stored with a short retention policy (the same applied to the appx stored as part of the CI). Whenever those builds are pushed to the MS Store, their PDB's also go to the repository wiki as compressed tar balls, with the intent of allowing retrieving from the wiki the always up-to-date set of PDB's. 

Later, during a release, we can take those artifacts from the wiki and upload them as assets for longer retention.

[This run](https://github.com/ubuntu/WSL/actions/runs/2786577116) successfullty collected the PDB files and uploaded them together with the appx, as well as the wiki. After that I changed the part that pushes to the wiki to be skipped if there was no upload to the store. [This run](https://github.com/ubuntu/WSL/actions/runs/2786706011) is the most updated, skipping the wiki part.

In order to certify that the PDB's are realy inside the wiki, one can `git clone https://github.com/ubuntu/WSL.wiki.git` and check the contents of the `debug-databases` directory. It should contain one tarball per Ubuntu app from which we collected PDBs during build. Those tarballs won't show up in GitHub website, since they are not text files.

We need to merge https://github.com/ubuntu/ubuntu-wsl-splash/pull/27 (and refresh the submodule of course) in order to see the PDB's from the ubuntu-wsl-splash Flutter app. We should also push a new version of Ubuntu 22.04 whenever all those changes are in `main` in order to have a collection of PDB's matching the exact compilation of the app pushed to the store.

### DIFF
--- a/.github/workflows/build-wsl.yaml
+++ b/.github/workflows/build-wsl.yaml
@@ -167,6 +167,28 @@ jobs:
       - name: Build Bundle
         working-directory: ${{ env.workDir }}
         run:  msbuild .\DistroLauncher.sln /t:Build /m /nr:false /p:Configuration=Release /p:AppxBundle=Always /p:AppxBundlePlatforms="${{ env.AppxBundlePlatforms }}" /p:UapAppxPackageBuildMode=${{ env.UapAppxPackageBuildMode }} -verbosity:normal
+      - name: Collect the PDB files
+        working-directory: ${{ env.workDir }}
+        shell: bash
+        run: |
+          set -eu
+          
+          # Launcher PDBs
+
+          for arch in "x64" "ARM64"; do
+            collectTo="debug-database-${{ matrix.AppID }}/launcher/$arch"
+            mkdir -p "$collectTo"
+            find "$arch/Release/" -name "*.pdb" -exec cp '{}' "$collectTo" \;
+          done
+
+          # Flutter PDBs
+
+          # Move this to the loop once Flutter supports ARM64.
+          collectTo="debug-database-${{ matrix.AppID }}/oobe/x64/"
+          mkdir -p "$collectTo"
+          find . -name "flutter_windows.dll.pdb" -exec cp '{}' "$collectTo" \;
+          find . -wholename "*ubuntu-*Release*.pdb" -exec cp '{}' "$collectTo" \;
+
       - name: Allow downloading sideload appxbundle
         uses: actions/upload-artifact@v2
         with:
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell (approved):
:rocket:

.github/workflows/build-wsl.yaml
```diff
@@ -167,6 +167,35 @@ jobs:
...
+
+          # Launcher PDBs
+
+          for arch in "x64" "ARM64"; do
+            collectTo="debug-database-${{ matrix.AppID }}/launcher/$arch/"
+            mkdir -p "$collectTo"
+
+            findDir="ARM64/Release/"
+            if [ $arch == "x64" ]; then
+              findDir="ARM64/Release/DistroLauncher-Appx/x64/"
```
toktamis, line 184: Can we double check if this path correct?

CarlosNihelton: One can search in the build-wsl workflow run logs for the compiler invocation with the `/PDB:`. That surprised me at first, the path is different from when I build the appx from the IDE.

toktamis: This is right. 2022-08-03T10:38:37.9059925Z          C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.29.30133\bin\HostX86\x64\link.exe /ERRORREPORT:QUEUE /OUT:"C:\Temp\builddir\ARM64\Release\DistroLauncher-Appx\x64\launcher.exe" /INCREMENTAL /ILK:"x64\Release\launcher.ilk" /NOLOGO onecore.lib /MANIFEST /MANIFESTUAC:"level='asInvoker' uiAccess='false'" /manifest:embed /DEBUG /PDB:"**C:\Temp\builddir\ARM64\Release\DistroLauncher-Appx\x64\launcher.pdb**" /SUBSYSTEM:CONSOLE /OPT:REF /OPT:ICF /LTCG:incremental /LTCGOUT:"x64\Release\launcher.iobj" /TLBID:1 /DYNAMICBASE /NXCOMPAT /IMPLIB:"C:\Temp\builddir\ARM64\Release\DistroLauncher-Appx\x64\launcher.lib" /MACHINE:X64 x64\Release\DistroLauncher.res

toktamis (approved):
Good work!

PR #267, opened 2022-08-15
Considers "Current Stable Release" as Supported for LTS's.

CarlosNihelton:
Releases information fetched from Launchpad API during the CI workflows currently looks like:

```
ver.	m.	code	full version	Status				Activ.	Sup.	LTS		Opened			Release			Milestone
22.10    0    kinetic    Ubuntu 22.10    Active Development    True    False    False    2022-04-26            
22.04    1    jammy    Ubuntu 22.04.1 LTS    Current Stable Release    True    True    True    2021-10-15    2022-04-21    2022-08-11    2023-02-09
20.04    4    focal    Ubuntu 20.04.4 LTS    Supported    True    True    True    2019-02-18    2020-04-23    2022-02-24    2022-09-01
18.04    5    bionic    Ubuntu 18.04.5 LTS    Supported    True    True    True    2017-10-04    2018-04-26    2020-08-06    
16.04    6    xenial    Ubuntu 16.04.6 LTS    Supported    True    True    True    2015-10-19    2016-04-21    2019-02-28    
14.04    6    trusty    Ubuntu 14.04.6 LTS    Supported    True    True    True    2012-10-01    2014-04-17    2019-03-07  
```

During the period of time between the LTS point-release and the next non-LTS release, the current LTS will appear as `Current Stable Release` instead of `Supported`. Thus, that version will only be assigned to the default app when the next non-LTS gets released.

```
ver.	m.	code	full version	Status
22.04    1    jammy    Ubuntu 22.04.1 LTS    Current Stable Release
```

It seems unfair to have to wait until a non-LTS release to be able to upgrade the default app to the latest LTS which already reached the point-release. For the Ubuntu Desktop, the point-release means it's time for the users running the previous LTS to upgrade. For Ubuntu WSL it means the same and should also mean that new users installing the default app should get the current LTS.

This PR changes one check that happens later in the loop that parses that information acquired from Launchpad, in which we set the default app ("Ubuntu") to the latest `Supported` version. We now allow it to also accept the `Current Stable Release`.

> That could have a side effect of allowing non-LTS versions to be assigned to the default app, but it won't happen because earlier in the loop we fast-forward the loop if the current line being parsed is a non-LTS release.

https://github.com/ubuntu/WSL/blob/b92552fbda202b184a10752d3c065dc60d1848db/wsl-builder/common/releasesinfo.go#L79-L82


CarlosNihelton:
PR #266 is a proof of this concept. It was automatically generated from CI from this branch. In fact I expected that #265 would be exactly that. The difference between them lies on the `Current Stable Release` behavior mentioned above.

CarlosNihelton:
When reviewing this PR, concentrate your efforts on commit e63efb11ae85c7e8f048fcf0adb59aa8b19d7dd9 . The others are automatically generated by the CI as an effect of that one.

toktamis (approved):

PR #269, opened 2022-08-29
Move the 'flutter build' commands from CI into MSBuild

CarlosNihelton:
The CI workflow is meant to provide the resources and execute the build system. Having it to know which Flutter project to builder (it could be either the current slide show or the new OOBE coming soon) and build it before invoking the actual build system that will output the appx bundle is a bad smell.

Instead we can let the build system (MSBuild) to handle building whatever dependencies the appx might have, only requiring the CI workflow to know that it must provide the Flutter command and that’s it.

This PR applies some renaming to turn it more subproject-agnostic in terms of whether we build the slide show or the OOBE application, as well as removing the build commands from CI workflow YAML files, leaving that to the actual build system (MSBuild) to handle.


CarlosNihelton:
[This run](https://github.com/ubuntu/WSL/actions/runs/2950177395) shows that we won't stop bundling the Flutter artifacts into the x64 part of the bundle.

image

EduardGomezEscandell (approved):

PR #271, opened 2022-09-13
Brings the OOBE to this repository

CarlosNihelton:
This PR brings `canonical/ubuntu-desktop-installer` as a submodule with the aim of building the Windows entry point of the package `ubuntu_wsl_setup`, which is our OOBE targeting the Windows platform.

Since, for now at least, it only supports Kinetic, we have to place two overrides in `meta/UbuntuPreview/src`:
- DistroLauncher-Apppx/Directory.Build.props - takes care of replacing the splash application by the ubuntu_wsl_setup.
- DistroLauncher/Preprocessor.props - `#define WIN_OOBE_ENABLED` for Ubuntu-Preview only, as originally discussed in #256 .

With those changes in place the build system is able to bundle the OOBE on the Ubuntu-Preview appx and the launcher knows which application strategy to use to start the slide show and the installer GUI.

Once this feature gets backported to Jammy, we will be able to safely place those in the default build tree, thus making them no more specific for Ubuntu-Preview only.


CarlosNihelton:
Notice the play with the CI: I wrote manually files in `/meta/UbuntuPreview/src/` and CI generated their equivalents on `/meta/UbuntuPreview/generated/`. In this case they are equal, but files in `src/` may have parameters that `wsl-builder` understands and interpolates during code generation.

EduardGomezEscandell (approved):
In the future it would be nice to not have the generated files in the repo, and have them be generated when building only, so as to avoid duplicates. Anyhow, this has nothing to do with this PR.

Ship it 🚀

CarlosNihelton:
Just wanted to register that with those changes in place, the Ubuntu-Preview appbundle built by [this run](https://github.com/ubuntu/WSL/actions/runs/3045275376) looks like:

x64 (notice the `ubuntu_wsl_setup` instead of splash, plus a couple of `dll`s):
image

ARM64:
image

PR #277, opened 2022-09-28
Closes the handle only at destructor

CarlosNihelton:
We previously tracked whether the event was set or not by assigning `nullptr` and to avoid leaks we close the handle right after setting it.

To preserve the semantics of being able to set it only once, we now track the set state using a boolean.

Deferring closing the handle to the destructor of this object allows for an OOBE GUI late in the initialization process to still find the event already set. Otherwise, closing the handle much earlier than any client would be able to check it, would allow the kernel to destroy the object.

This was never an issue because the OOBE inits very fast. But with the upcoming end to end test this would be a major block: the launcher would never be effective in communicating the registration completion event.

### DIFF
--- a/DistroLauncher/SetOnceNamedEvent.cpp
+++ b/DistroLauncher/SetOnceNamedEvent.cpp
@@ -26,6 +26,7 @@ namespace Win32Utils
         HANDLE handle = CreateEvent(nullptr, TRUE, FALSE, name);
         if (handle != nullptr) {
             event = handle;
+            setOnce = false;
         }
     }
 
@@ -34,12 +35,13 @@ namespace Win32Utils
         if (event != nullptr) {
             CloseHandle(event);
             event = nullptr;
+            setOnce = false;
         }
     }
 
     bool SetOnceNamedEvent::set() noexcept
     {
-        if (event == nullptr) {
+        if (setOnce == true) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


DistroLauncher/SetOnceNamedEvent.cpp
```diff
@@ -26,6 +26,7 @@ namespace Win32Utils
         HANDLE handle = CreateEvent(nullptr, TRUE, FALSE, name);
         if (handle != nullptr) {
             event = handle;
+            setOnce = false;
         }
```
EduardGomezEscandell, line 30: This `if` block is unnecessary, the alternative does the same.

You could assign the event to the handle directly.

DistroLauncher/SetOnceNamedEvent.cpp
```diff
@@ -34,31 +35,31 @@ namespace Win32Utils
         if (event != nullptr) {
             CloseHandle(event);
             event = nullptr;
+            setOnce = false;
```
EduardGomezEscandell, line 38: No need to set variables in the destructor

EduardGomezEscandell (changes requested):
Small details, plus clang-tidy's advise.

DistroLauncher/SetOnceNamedEvent.cpp
CarlosNihelton: Indeed 🙈

DistroLauncher/SetOnceNamedEvent.cpp
CarlosNihelton: Agreed

DistroLauncher/SetOnceNamedEvent.cpp
```diff
@@ -34,31 +35,31 @@ namespace Win32Utils
...
     SetOnceNamedEvent::SetOnceNamedEvent(SetOnceNamedEvent&& other) noexcept :
-        event{std::exchange(other.event, nullptr)}
+        event{std::exchange(other.event, nullptr)}, setOnce{std::exchange(other.setOnce, false)}
     {
     }
 
     bool SetOnceNamedEvent::isValid() const noexcept
     {
-        return event != nullptr;
+        return event != nullptr && setOnce == false;
```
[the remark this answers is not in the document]
CarlosNihelton: Fixed

EduardGomezEscandell (commented):
Great. Also +1 on naming it `alreadySet`.

CarlosNihelton:
> Great. Also +1 on naming it `alreadySet`.

I should avoid naming things in the end of the day... :D

EduardGomezEscandell (approved):

PR #278, opened 2022-09-29
Ensures the child process has at least one thread available for input

CarlosNihelton:
This doesn't seem useful if the child is a window application. But showed useful when dealing with the PoC e2e test.

This approach is quite controversial though. See the following links for the downside of doing such.

https://devblogs.microsoft.com/oldnewthing/20100325-00/?p=14493

https://devblogs.microsoft.com/oldnewthing/20100326-00/?p=14483

Still seems as harmful as doing nothing (the previous behavior) for GUI child processes.

### DIFF
--- a/DistroLauncher/ChildProcess.cpp
+++ b/DistroLauncher/ChildProcess.cpp
@@ -60,6 +60,7 @@ namespace Oobe
                                  &procInfo);                // output: PROCESS_INFORMATION
         auto ok = res != 0 && procInfo.hProcess != nullptr; // success
         if (ok) {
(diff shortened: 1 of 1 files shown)


toktamis (approved):
The blog is discussing its call more than once to determine process status. As the the title describes "WaitForInputIdle should really be called WaitForProcessStartupComplete
" the main purpose is capturing if the Process startup complete or not. This means at least one thread will be ready. This meets your expectation. You don't need to check the status more than once so the controversial part is not related to this change.

PR #280, opened 2022-10-04
Fix flutter build CI

CarlosNihelton:
There was a breaking change in the package `flutter_html` that motivated the `ubuntu_desktop_installer` to pin a very specific version of that package. See https://github.com/canonical/ubuntu-desktop-installer/pull/1152. That is the reason for the current break in our CI build. Thus, we need to update the UDI submodule. 

I also took the freedom to unpin the flutter-action version from our CI.

[This run](https://github.com/ubuntu/WSL/actions/runs/3184611222) proves this work.


EduardGomezEscandell (approved):
Great!

PR #281, opened 2022-10-04
End to end testing architecture

CarlosNihelton:
This implements the Go side of the end-to-end testing workflow.

We have two Go modules grouped in a workspace:
- `ui-driver` replaces the OOBE binary, matching its command line and selecting the appropriate Dart test case to drive the real OOBE GUI. It also contains a MSBuild `.props` file that enables injecting the `ui-driver.exe` binary inside the appx at build time.
- `test-runner` implements the actual assertions based on which command line we invoke the distro launcher with. Notice that in this package we don't have a main function. It's meant to be invoked with `go test`, parameterized with launcher name (`ubuntupreview.exe` for example) and distro name (`Ubuntu-Preview`).

Feel free to push back or request further explanation on the implementation details.


e2e/test-runner/wsl_tester.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Runs the specified command and arguments by means of os.exec (no shell involved.)
+// Asserts that the process succeeds and returns its combined output (stdout + stderr).
+func (t *Tester) AssertOsCommand(name string, args ...string) string {
+	cmd := exec.Command(name, args...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		t.Fatalf("Failed to run the command: %s. Error %s", name, err)
+	}
```
EduardGomezEscandell, line 80: We may want to differentiate between:
- Failed to run the command (implies the command was not run for some reason)
- Command returned non-zero exit status

e2e/test-runner/runner_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+func TestBasicSetup(t *testing.T) {
+	// Wrapping testing.T to get the friendly additions to its API as declared in `wsl_tester.go`.
+	tester := WslTester(t)
+
+	// Invoke the launcher as if the user did: `<launcher>.exe install --ui=gui`
+	outStr := tester.AssertLauncherCommand("install --ui=gui") // without install the launcher will run bash.
+	if len(outStr) == 0 {
+		tester.Fatal("Nothing happened.")
```
EduardGomezEscandell, line 15:
```suggestion
		tester.Fatal("Failed to install the distro: No output produced.")
```
Clearer 😃

e2e/test-runner/wsl_tester.go
```diff
@@ -0,0 +1,82 @@
...
+			rootDir := os.Getenv("GITHUB_WORKSPACE")
+			path := filepath.Join(rootDir, clientLogPath)
+			clientLogContents, err := ioutil.ReadFile(path)
+			if err != nil {
+				tester.Fatalf("%s", err)
+			}
+			tester.Logf("%s", clientLogContents)
+		}
+		// unregister the instance.
+		tester.AssertOsCommand("wsl.exe", "--unregister", *distroName)
```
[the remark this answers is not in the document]
EduardGomezEscandell:
```Powershell
PS> wsl --unregister UbuntuDev.WslID.Dev
Unregistering.
There is an installation, uninstallation or conversion in progress for this distribution.

Error code: Wsl/Service/0x8000000d
```
After waiting for a second or so, I'm allowed to unregister so no problem. We don't need to wait, we can just shut it down.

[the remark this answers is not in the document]
EduardGomezEscandell: Hmm it's more complicated.
- Normal execution: You can simply unregister
- Certain broken installs: You can only unregister after ~1 second.
  - If you try to `wsl -t $DISTRO`, it doesn't find it
  - If you `wsl --shutdown`, then it is unregistered automatically and calling unregister afterwards fails.

e2e/test-runner/wsl_tester.go
```diff
@@ -0,0 +1,82 @@
...
+// Runs 'wsl.exe -d distroName' command line supplied as arguments to the wsl.exe command.
+// Asserts that the process succeeds and returns its combined output (stdout + stderr).
+func (t *Tester) AssertWslCommand(args ...string) string {
+	fullArgs := append([]string{"-d", *distroName}, args...)
+	return t.AssertOsCommand("wsl.exe", fullArgs...)
+}
+
+// Runs the command line supplied as arguments to the configured distro launcher on powershell.
+// Asserts that the process succeeds and returns its combined output (stdout + stderr).
+func (t *Tester) AssertLauncherCommand(fullCommand string) string {
```
EduardGomezEscandell, line 68: Could be a good idea to allow for an extra argument for an error explanation:
```go
t.AssertLauncherCommand("test -f /var/log/upgrade-policy-changed.log", "Launcher did not crate sentinel file after changing upgrade policy.")
```
As tests get more abundant and specific we'll find these messages useful to quickly triage errors.

Not married to the idea, so please provide your thoughts.

EduardGomezEscandell (commented):
As you see many of my comments are regarding clarity, particularly about why errors fail.

I become very frustrated when working with codebases where you have have to be intimately familiar with the entire code base to understand the most peripheral failed test.

I'm hoping to avoid this here, although I may be going overboard 😄.

EduardGomezEscandell:
Also in a followup PR I'd add a readme to the e2e folder, as it is not immediately evident that e2e means Integration tests, or how they are architected.

e2e/test-runner/runner_test.go
CarlosNihelton: Indeed.

e2e/test-runner/wsl_tester.go
```diff
@@ -0,0 +1,82 @@
...
+var launcherName = flag.String("launcher-name", defaultLauncherName, "WSL distro launcher under test.")
+var distroName = flag.String("distro-name", defaultDistroName, "WSL distro instance registered for testing.")
+
+// A testing wrapper with functions simplifying composing and running commands through os.Exec and checking their outputs.
+type Tester struct {
+	*testing.T
+}
+
+func checkTestEnvMinRequirements(t *testing.T) {
+	rootDir := os.Getenv("GITHUB_WORKSPACE")
```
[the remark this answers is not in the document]
CarlosNihelton: `$GITHUB_WORKSPACE ` is a well-known environment variable in the context of GitHub actions, set by the checkout action, which is usually the first step of any workflow. Calling it something else is not a problem but it requires the CI workflow to create a new environment variable with the same value as this one. I'd expect the check (and error message) below to be enough to clarify the situation when this variable is not set, since it aborts the test in the very beginning.

Even though, I'm not exactly disagreeing with you.

e2e/test-runner/wsl_tester.go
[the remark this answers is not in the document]
CarlosNihelton: Good point. I think shutting down WSL is the best way. Errors are irrelevant at this point.

e2e/test-runner/wsl_tester.go
CarlosNihelton: Well, the semantics of an error depends on which command is run. If a command exits 0 and produces no output when you expect for an output is in fact an error. So, to follow this idea we'd need to also pass a predicate function informing whether the command output means an error or not.

Still, the string can be supplied as a customization point for the cases when the command exits non-zero. Again, whether this is the only error situation possible or not depends on the command and expectations.

e2e/test-runner/wsl_tester.go
CarlosNihelton: Not sure if the added complexity is worth. We'd need to invoke `cmd.Start()` and check for errors that mean the process failed to start (command not found in path, for instance). Then we'd need to create a goroutine to keep reading stdout and stderr pipes to put into the string we want to return from this function, otherwise we'd lose the outputs. Then we finish calling `cmd.Wait()`, which blocks until the process finishes, so we can check if it exited 0. Calling `Wait()` closes the stdout and stderr pipes, thus the goroutines would be required.

What do you think?

e2e/test-runner/wsl_tester.go
CarlosNihelton: Also, it hurts the API a little bit, because the current way allows passing a variadic number of strings composing the command line arguments. Allowing an error message to be passed require being clever on how to differentiate that string from the rest and not passing it to the command line. At a first glance it seems we'd need to add a type for that.

e2e/test-runner/wsl_tester.go
EduardGomezEscandell: Yeah you're right, it's overkill. I'd keep it as-is, if more granularity is needed then we can manually call the command instead of using this function.

CarlosNihelton:
> Also in a followup PR I'd add a short readme to the e2e folder, as it is not immediately evident that e2e means Integration tests, or how they are architected.

I'm adding the README now. In the follow up PR I'll have it enhanced as I plan to push also a script that is the basis for the CI workflow as well.

EduardGomezEscandell (approved):
Great, this will really streamline our release process and make changes safer.

I really like the solution of having a constants module 😃

PR #286, opened 2022-10-05
Integrates the end-to-end testing with the distro launcher

CarlosNihelton:
This seems bigger than it really is. Let's go by parts:

- ubuntu-desktop-installer submodule is updated to latest main so we can incorporate the automation recently merged there. See PR https://github.com/canonical/ubuntu-desktop-installer/pull/1173
- A small change in `meta\UbuntuPreview\Preprocessor.props` allows defining `OOBE_E2E_TESTING` through msbuild CLI. That will only make sense in the context of running the end-to-end tests.
- AppConfig struct and a function to provide a global static constexpr instance carries the pieces of data that change between production and e2e testing.
- Fixes and enhancements on the launcher code revealed by the e2e testing experimentation. (2ceb9e3b640bbe641e8c3f8f8400f62a910515ef, a3e0f476881acfca9293b9a631449d5f740d3e41 and 57f465b0cbbb1b1eddc84456916cb585bc0f7754)

We are almost done. With those changes in place it's possible to generate an appx ready for end-to-end testing by following the recipe below on a Developer Powershell:

```powershell

 go build .\wsl-builder\prepare-build
 .\prepare-build Ubuntu-Preview ...
cp e2e\ui-driver\ui-driver.props DistroLauncher-Appx\
msbuild .\DistroLauncher.sln /t:Build /m /nr:false /p:Configuration=Debug /p:AppxBundle=Always /p:AppxBundlePlatforms="x64" /p:UapAppxPackageBuildMode=SideloadOnly -verbosity:normal /p:OOBE_E2E_TESTING=1
.\AppPackages\Ubuntu\Ubuntu_<version>_Debug_Test\Install.ps1 -Force
$env:LAUNCHER_REPO_ROOT=$(pwd).Path
cd e2e
go test .\test-runner --distro-name Ubuntu-Preview --launcher-name ubuntupreview.exe

```

This will soon be turned into a script, on another PR, once experimentation with the parameters get more mature.

### DIFF
--- a/DistroLauncher/AppConfig.cpp
+++ b/DistroLauncher/AppConfig.cpp
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2022 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include "stdafx.h"
+
+namespace Oobe
+{
+    AppConfig appConfig()
+    {
+        // Initialized only once at compile-time and forever immutable.
+        static constexpr AppConfig instance =
+#ifdef OOBE_E2E_TESTING
+          AppConfig{L"ui-driver.exe", true, true};
+#else
+          AppConfig{L"ubuntu_wsl_setup.exe", false, false};
+#endif
+        return instance;
+    }
+}
--- a/DistroLauncher/AppConfig.h
+++ b/DistroLauncher/AppConfig.h
@@ -1,0 +1,35 @@
+#pragma once
+/*
+ * Copyright (C) 2022 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace Oobe
+{
+    struct AppConfig
+    {
+        // The file name (including extension) of the executable that must be invoked as the OOBE.
+        wchar_t* appName;
+
+        // Whether or not the fallback method should be skipped on OOBE failure.
+        bool mustSkipFallback;
+
+        // Whether the child process requires its own console or not.
+        bool requiresNewConsole;
+    };
+
+    // Returns the global constant instance of the AppConfig.
+    AppConfig appConfig();
+}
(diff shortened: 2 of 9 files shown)


DistroLauncher/AppConfig.cpp
```diff
@@ -0,0 +1,33 @@
...
+namespace Oobe
+{
+    AppConfig appConfig()
+    {
+        // Initialized only once at compile-time and forever immutable.
+        static constexpr AppConfig instance =
+#ifdef OOBE_E2E_TESTING
+          AppConfig{L"ui-driver.exe", true, true};
+#else
+          AppConfig{L"ubuntu_wsl_setup.exe", false, false};
```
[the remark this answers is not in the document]
CarlosNihelton: Solved!

DistroLauncher/AppConfig.h
```diff
@@ -0,0 +1,35 @@
...
+
+        // Whether or not the fallback method should be skipped on OOBE failure.
+        bool mustSkipFallback;
+
+        // Whether the child process requires its own console or not.
+        bool requiresNewConsole;
+    };
+
+    // Returns the global constant instance of the AppConfig.
+    AppConfig appConfig();
```
EduardGomezEscandell, line 34:
```suggestion
    struct AppConfig
    {
        // The file name (including extension) of the executable that must be invoked as the OOBE.
        constexpr std::wstring_view appName;

        // Whether or not the fallback method should be skipped on OOBE failure.
        constexpr bool mustSkipFallback;

        // Whether the child process requires its own console or not.
        constexpr bool requiresNewConsole;
    };

    // Returns the global constant instance of the AppConfig.
    constexpr AppConfig appConfig();
```
We're not very performance-blocked but might as well. Might even want to move the definition of `appConfig` to this header file to go all-in and make it trully constexpr.

DistroLauncher/AppConfig.h
```diff
@@ -0,0 +1,35 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace Oobe
+{
+    struct AppConfig
+    {
+        // The file name (including extension) of the executable that must be invoked as the OOBE.
+        const wchar_t* appName;
```
EduardGomezEscandell, line 24: IMO we should be a bit more specific. `OobeExecutableName` or something of the sort. `AppConfig.appName` would appear to refer to the launcher appx's name (e.g. `Ubuntu (Preview)` or `Ubuntu 22.04.1 LTS`).

DistroLauncher/Application.h
```diff
@@ -84,6 +84,9 @@ namespace Oobe
             if (FAILED(hr) && hr != E_NOTIMPL && hr != E_INVALIDARG) {
                 wprintf(L"Installer did not complete successfully.\n");
                 Helpers::PrintErrorMessage(hr);
+                if (appConfig().mustSkipFallback) {
+                    exit(123);
```
EduardGomezEscandell, line 88: Why `123`?

I'd rather have some variable at the top of the file, something like `constexpr int EXIT_OOBE_NO_FALLBACK = 123`.

Ideally we'd have a file defining all possible exit values, but since it's only this one I think that'd be overkill.

CarlosNihelton: Agreed.

DistroLauncher/AppConfig.h
CarlosNihelton: Agreed.

DistroLauncher/AppConfig.h
CarlosNihelton: Not a big fan of making all data members const(expr) when we can make the whole object const(expr) from the outside. Not in this case, but in other cases that could hurt testability (thinking of the const-ness).

Also, I'd prefer to avoid showing #ifdef's in the header files. Following this premise require the function to be defined in the cpp, thus it would never be truly constexpr.

But the function being a singleton provider, it makes sense to make it return a const ref instead.

CarlosNihelton:
Jammy was moved to the new pipeline. I didn't check yet the results, but it is the reason why the CI is failing to build a package.

DistroLauncher/AppConfig.h
EduardGomezEscandell: Makes sense to me. However I believe that moving it to the sourcefile makes it not be constexpr (constexpr definitions need be in the same compilation unit). However this is very simple so LTO may even optimize it for us.

EduardGomezEscandell (approved):
👍

PR #292, opened 2022-10-10
Adding e2e testing relating to config 

EduardGomezEscandell:
The launcher modifies some config files in WSL. These tests verify this happens.

Feature commits:
- [96dc60f](https://github.com/ubuntu/WSL/pull/292/commits/96dc60fed90c6e378b7a75c28bef92b7b8a8c180) tests that systemd is enabled and that the fixes the launcher applies to to sysusers service work.
- [8acfd93](https://github.com/ubuntu/WSL/pull/292/commits/8acfd9380b3c1698bbf99fa154d86940e15e11ae) tests that the upgrade policy is set according to the distro name
- [5f1d754](https://github.com/ubuntu/WSL/pull/292/commits/5f1d754a91c292480008406d5571fabb2ef7befc) tests that we used the right rootfs link. This test will have to be updated every time there is new release. I didn't want to automate it because the point of this test is to test this very automation.

Cleanup commits:
- [7cf259e](https://github.com/ubuntu/WSL/pull/292/commits/7cf259e33d6b861e82a6feda516e7283cd61516f). This commit is best viewed without whitespace diffs. I enclosed each set of assertions in their own scope to avoid bleeding variables from one test to the next one. This commit makes big enough changes that merging the latter cleanup commits with the previous ones is too much of a hassle.

Later commits are bugfixes and smaller changes.

### DIFF
--- a/e2e/test-runner/go.mod
+++ b/e2e/test-runner/go.mod
@@ -2,8 +2,8 @@ module github.com/ubuntu/wsl/e2e/test-runner
 
 go 1.18
 
-require github.com/ubuntu/wsl/e2e/constants v0.0.0
-
-require gopkg.in/ini.v1 v1.67.0 // indirect
-
+require (
+    github.com/ubuntu/wsl/e2e/constants v0.0.0
+    gopkg.in/ini.v1 v1.67.0 // indirect
+)
 replace github.com/ubuntu/wsl/e2e/constants => ../constants
--- a/e2e/test-runner/go.mod
+++ b/e2e/test-runner/go.mod
@@ -3,4 +3,7 @@ module github.com/ubuntu/wsl/e2e/test-runner
 go 1.18
 
 require github.com/ubuntu/wsl/e2e/constants v0.0.0
+
+require gopkg.in/ini.v1 v1.67.0 // indirect
+
 replace github.com/ubuntu/wsl/e2e/constants => ../constants
--- a/e2e/test-runner/runner_test.go
+++ b/e2e/test-runner/runner_test.go
@@ -1,9 +1,21 @@
 package test_runner
 
 import (
+	"gopkg.in/ini.v1"
 	"strings"
 	"testing"
 )
+
+func expectedUpgradePolicy() string {
+	if *distroName == "Ubuntu" {
+		return "lts"
+	}
+	if strings.HasPrefix(*distroName, "Ubuntu") && strings.HasSuffix(*distroName, "LTS") {
+		return "never"
+	}
+	// Preview and Dev
+	return "normal"
+}
 
 func TestBasicSetup(t *testing.T) {
 	// Wrapping testing.T to get the friendly additions to its API as declared in `wsl_tester.go`.
@@ -45,4 +57,36 @@ func TestBasicSetup(t *testing.T) {
 
 	// Sysusers service fix
 	tester.AssertWslCommand("systemctl", "status", "systemd-sysusers.service")
+
+	// ---------------- Reboot with launcher -----------------
+	tester.AssertOsCommand("wsl.exe", "-t", *distroName)
+	tester.AssertLauncherCommand("run echo Hello")
+
+	// Upgrade policy
+	outputStr = tester.AssertWslCommand("cat", "/etc/update-manager/release-upgrades")
+	cfg, err := ini.Load([]byte(outputStr))
+	if err != nil {
+		tester.Logf("Contents of /etc/update-manager/release-upgrades:\n%s", outputStr)
+		tester.Fatal("Failed to parse ini file")
+	}
+
+	val := cfg.Section("DEFAULT").Key("Prompt").String()
+	if val != expectedUpgradePolicy() {
+		tester.Logf("Contents of /etc/update-manager/release-upgrades:\n%s", outputStr)
+		tester.Logf("Parsed policy: %s", val)
+		tester.Logf("Expected policy: %s", expectedUpgradePolicy())
+		tester.Fatal("Wrong upgrade policy")
+	}
+
+	release_upgrades_date := tester.AssertWslCommand("date", "-r", "/etc/update-manager/release-upgrades")
+
+	// ---------------- Reboot with launcher -----------------
+	tester.AssertOsCommand("wsl.exe", "-t", *distroName)
+	tester.AssertLauncherCommand("run echo Hello")
+
+	new_date := tester.AssertWslCommand("date", "-r", "/etc/update-manager/release-upgrades")
+	if release_upgrades_date != new_date {
+		tester.Logf("Launcher is modifying release upgrade file more than once")
+	}
+
 }
--- a/e2e/test-runner/runner_test.go
+++ b/e2e/test-runner/runner_test.go
@@ -23,7 +23,7 @@ func TestBasicSetup(t *testing.T) {
 	tester := WslTester(t)
 
 	// Invoke the launcher as if the user did: `<launcher>.exe install --ui=gui`
-	outStr := tester.AssertLauncherCommand("install --root --ui=none") // without install the launcher will run bash.
+	outStr := tester.AssertLauncherCommand("install --ui=gui") // without install the launcher will run bash.
 	if len(outStr) == 0 {
 		tester.Fatal("Failed to install the distro: No output produced.")
 	}
--- a/e2e/test-runner/runner_test.go
+++ b/e2e/test-runner/runner_test.go
@@ -79,15 +79,17 @@ func assertCorrectReleaseRootfs(tester *Tester) {
 
 // Ensures systemd was enabled.
 func assertSystemdEnabled(tester *Tester) {
+	tester.AssertWslCommand("busctl", "list", "--no-pager")
+
 	getSystemdStatus := func() string {
-		outputStr := tester.AssertWslCommand("bash", "-ec", "systemctl is-system-running || exit 0")
+		outputStr := tester.AssertWslCommand("bash", "-ec", "systemctl is-system-running --wait || exit 0")
 		return strings.TrimSpace(outputStr)
 	}
 
 	status := getSystemdStatus()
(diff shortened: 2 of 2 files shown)


e2e/test-runner/runner_test.go
```diff
@@ -1,35 +1,162 @@
...
+	return "normal"
+}
+
 func TestBasicSetup(t *testing.T) {
 	// Wrapping testing.T to get the friendly additions to its API as declared in `wsl_tester.go`.
 	tester := WslTester(t)
 
 	// Invoke the launcher as if the user did: `<launcher>.exe install --ui=gui`
-	outStr := tester.AssertLauncherCommand("install --ui=gui") // without install the launcher will run bash.
+	outStr := tester.AssertLauncherCommand("install --root --ui=none") // without install the launcher will run bash.
```
EduardGomezEscandell, line 26:
```suggestion
	outStr := tester.AssertLauncherCommand("install --ui=gui") // without install the launcher will run bash.
```
This debug code slipped through :(

e2e/test-runner/go.mod
```diff
@@ -3,4 +3,7 @@ module github.com/ubuntu/wsl/e2e/test-runner
 go 1.18
 
 require github.com/ubuntu/wsl/e2e/constants v0.0.0
+
+require gopkg.in/ini.v1 v1.67.0 // indirect
+
```
CarlosNihelton, line 8: We can do it in a single require clause:

```suggestion
require (
    github.com/ubuntu/wsl/e2e/constants v0.0.0
    gopkg.in/ini.v1 v1.67.0 // indirect
)
```

e2e/test-runner/runner_test.go
```diff
@@ -1,10 +1,23 @@
...
 
 import (
+	"gopkg.in/ini.v1"
 	"strings"
 	"testing"
+	"time"
 )
 
+func expectedUpgradePolicy() string {
+	if *distroName == "Ubuntu" {
```
CarlosNihelton, line 11: I'd recommend passing the pointer to the distroName as an argument, since `expectedUpgradePolicy()` is meant to be called in the middle of a higher-level function. We may change the global variable name later in the future, or the way it's set (causing it to even not be a global anymore).

e2e/test-runner/runner_test.go
```diff
@@ -16,20 +29,134 @@ func TestBasicSetup(t *testing.T) {
 	}
 
 	// After completing the setup, the OOBE must have created a new user and the distro launcher must have set it as the default.
-	outputStr := tester.AssertWslCommand("whoami")
-	notRoot := strings.Fields(outputStr)[0]
-	if notRoot == "root" {
-		tester.Logf("%s", outputStr)
-		tester.Fatal("Default user should not be root.")
+	{
```
CarlosNihelton, line 32: As you realized those blocks could have their own scope, I think we should actually put names on them and call them as functions. Maybe we can reuse some of those functions in other test scenarios beyond the basic setup.

e2e/test-runner/runner_test.go
```diff
@@ -16,20 +29,134 @@ func TestBasicSetup(t *testing.T) {
...
+			if *distroName == "Ubuntu22.04LTS" {
+				return "Ubuntu 22.04.1 LTS"
+			}
+			if *distroName == "Ubuntu20.04LTS" {
+				return "Ubuntu 20.04.5 LTS"
+			}
+			if *distroName == "Ubuntu18.04LTS" {
+				return "Ubuntu 18.04.6 LTS"
+			}
+			return ""
```
CarlosNihelton, line 69: This closure requires change whenever a point release of any LTS or whenever the development release gets published or changed into the following cycle. Seems excessive work. We should consider making it more stable, like returning only the series (for example: `18,.04`, `20.04`) and avoid assertions at all if `*distroName == "Ubuntu-Preview" || *distroName == "UbuntuDev.WslID.Dev"`. Or, if we want to be really strict with the release version to the point release, we should consider making it a parameter passed by the CI workflow. What do you think?

e2e/test-runner/runner_test.go
```diff
@@ -16,20 +29,134 @@ func TestBasicSetup(t *testing.T) {
...
+		}()
+		if len(expectedRelease) == 0 {
+			tester.Logf("Unknown Ubuntu release corresponding to distro name '%s'", *distroName)
+			tester.Fatal("Unexpected value provided via --distro-name")
+		}
+		outputStr := tester.AssertWslCommand("cat", "/etc/os-release")
```
CarlosNihelton, line 75: Honest question: why this is preferred over `lsb_release -d` ?

```
cjam@Zero01:~$ lsb_release -d
Description:    Ubuntu 22.04.1 LTS
```

e2e/test-runner/runner_test.go
```diff
@@ -16,20 +29,134 @@ func TestBasicSetup(t *testing.T) {
...
+		getSystemdStatus := func() string {
+			outputStr := tester.AssertWslCommand("bash", "-ec", "systemctl is-system-running || exit 0")
+			return strings.TrimSpace(outputStr)
+		}
+
+		status := getSystemdStatus()
+		for status == "starting" {
+			time.Sleep(1 * time.Second)
+			status = getSystemdStatus()
+		}
```
CarlosNihelton, line 107: If we only want to check whether systemd was enabled or not we could accept `starting` as a valid answer and move on.

If, instead, we want to make sure systemd is really running, than that's fine. An alternative would be running once the command `systemctl is-system-running --wait` (since v240). No (raw) loops. 😅 . Both approaches are susceptible to failure if D-Bus daemon itself has not yet started, because the systemctl command would fail to connect to D-Bus. I don't know how feasible that is to happen on CI, but we should consider that path otherwise we may turn into brittle tests for unknown reasons. We can run `busctl list --no-pager` first to ensure D-Bus is running before attempting to block on systemd just in case. That command will fail fast if D-Bus is not running.

All tests I did with WSL, though, seem to reveal that WSL waits on systemd before running the systemctl command. I don't think we should rely heavily on that, though.

CarlosNihelton (changes requested):
I think we have room for improvement here.

e2e/test-runner/runner_test.go
EduardGomezEscandell: Fair enough. Done.

e2e/test-runner/runner_test.go
EduardGomezEscandell: This test I created as a proposal more than as a done thing. When a new release comes out we always manually check if the correct dot version is downloaded. This semi-automated that. I'm not sure how we could automate it without circular reasoning (e.g: if the pipeline does not have Ubuntu 22.04.2 when time is due, how can we know without manually checking?).

EduardGomezEscandell: I relaxed the assertion to only check for major and minor without dot release, but I believe it is worth considering.

e2e/test-runner/runner_test.go
EduardGomezEscandell: This test I created as a proposal more than as a done thing. When a new release comes out we always manually check if the correct dot version is downloaded. This semi-automated that. I'm not sure how we could automate it without circular reasoning (e.g: if the pipeline does not have Ubuntu 22.04.2 when time is due, how can we know without manually checking?).

I relaxed the assertion to only check for major and minor without dot release, but I believe it is worth considering.

e2e/test-runner/runner_test.go
EduardGomezEscandell: Because I forgot to check for lsb_release flag options 😅. I changed it now.

e2e/test-runner/runner_test.go
EduardGomezEscandell: Better safe than sorry. I added both suggestions.

By using `systemctl is-system-running --wait` instead of directly accepting `starting` we also have the implicit check that it does not fail during startup.

e2e/test-runner/runner_test.go
```diff
@@ -1,35 +1,169 @@
...
+	outputStr := tester.AssertWslCommand("lsb_release", "-r")
+	cfg, err := ini.Load([]byte(outputStr))
+	if err != nil {
+		tester.Logf("Contents of /etc/os-release:\n%s", outputStr)
+		tester.Fatal("Failed to parse ini file")
+	}
+
+	release := cfg.Section(ini.DefaultSection).Key("Release")
+	if release.String() != expectedRelease {
+		tester.Logf("Contents of /etc/os-release:\n%s", outputStr)
```
CarlosNihelton, line 73: Outdated?

e2e/test-runner/runner_test.go
CarlosNihelton: Let's park that idea for the soon future. I think the way to go is having the CI passing which distro version is expected. CI would now by inquiring Launchpad. We already do that when the `build-wsl` workflow decides which versions to build. You'd notice that sometimes it builds more than just the Ubuntu-Preview.  This is a lot more complicated, so I'd rather we investigate that idea in a dedicated PR.

e2e/test-runner/runner_test.go
EduardGomezEscandell: Good catch

e2e/test-runner/runner_test.go
```diff
@@ -1,35 +1,169 @@
...
+		return "22.10" // Development version
+	}()
+	if len(expectedRelease) == 0 {
+		tester.Logf("Unknown Ubuntu release corresponding to distro name '%s'", *distroName)
+		tester.Fatal("Unexpected value provided via --distro-name")
+	}
+	outputStr := tester.AssertWslCommand("lsb_release", "-r")
+	cfg, err := ini.Load([]byte(outputStr))
+	if err != nil {
+		tester.Logf("Contents of /etc/os-release:\n%s", outputStr)
```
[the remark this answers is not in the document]
EduardGomezEscandell: Good catch

CarlosNihelton (approved):
LGTM!

Great job with the function split-up! It looks gorgeous!

PR #293, opened 2022-10-10
Adding test to validate most common installation procedure

EduardGomezEscandell:
## Motivation

`launcher.exe --hide-console` is the command line used by the Start Menu and by the Store app installer to invoke the distro launcher. That is most likely the preferred way users interact with the distro launcher during the initial setup. When the setup is complete the launcher runs the shell, which depends on user interaction to exit.

## Initial (bad) idea
In order to mimick this, we launch the aforementioned command in a forked process and monitor it from the original process by capturing its StdErr, StdOut and the distro status as shown by `wsl -l -v`. Note that I have not found a way to capture the text written to console by WSL's interactive session, as it is sent neither to StdOut nor StdErr.

*Result*:  WSL knows it's not running in a terminal and hence does not run the same way. For instance, it's not held open indefinitely. (This is also likely the reason I didn't capture any text: it is never displayed).

## Current approach
Open a new tab in Windows Terminal. This is basically the same as forking the process, with the added bonus of WSL detecting it is running in a terminal:
```PS
wt.exe --window 0 -d . powershell.exe -noExit launcher.exe --hide-console
#                                     ^~~~~~~
#                                     WSL never shuts down without this 🤷 
```

I have not found how to read from this console either.

### DIFF
--- a/e2e/test-runner/runner_test.go
+++ b/e2e/test-runner/runner_test.go
@@ -10,8 +10,10 @@ import (
 )
 
 func getDistroStatus(t *Tester) string {
+	// wsl -l -v outputs UTF-16 (See https://github.com/microsoft/WSL/issues/4607)
+	// We use $env:WSL_UTF8=1 to prevent this (Available from 0.64.0 onwards https://github.com/microsoft/WSL/releases/tag/0.64.0)
 	str := t.AssertOsCommand("powershell.exe", "-noninteractive", "-nologo", "-noprofile", "-command",
-		"(wsl -l -v) -replace '\\x00'")
+		"$env:WSL_UTF8=1 ; wsl -l -v")
 
 	for _, line := range strings.Split(str, "\n")[1:] {
 		columns := strings.Fields(line)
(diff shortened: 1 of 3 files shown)


e2e/test-runner/runner_test.go
```diff
@@ -33,3 +58,49 @@ func TestBasicSetup(t *testing.T) {
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	cmd.Stderr = &out
+
+	cmd.Start()
+	time.Sleep(1 * time.Second)
+
+	// Polling until status is no longer Installing
+	status := getDistroStatus(&tester)
+	for status == "Installing" {
```
EduardGomezEscandell, line 83: This Sleep(1 second) is a bit ugly but I couldn't think of a better option to avoid failing the loop condition. Suggestions are welcome.

EduardGomezEscandell: Solved!

EduardGomezEscandell:
There is currently a race where sometimes Stdout is not flushed before the distro status switches to Stopped.

CarlosNihelton (approved):
I'm not in love with this approach, but we can think of ways to improve it at next.

PR #295, opened 2022-10-13
e2e: Removing old log before running tests

EduardGomezEscandell:
When a test fails early during installation, the symlink to the ubuntu-desktop-installer logs points to the previous test's log. This can be confusing.

This PR erases the symlink (not the log itself) during setup.


CarlosNihelton (approved):
LGTM

PR #296, opened 2022-10-17
Updates UDI after fixing ApplyingChangesPage loop and crash

CarlosNihelton:
When working with @EduardGomezEscandell to get the tests running on his machine we noticed two issues related to the ApplyingChangesPage that could lead to:

1. An infinite loop with the GUI waiting for the server to end when the server already ended its job. https://github.com/canonical/ubuntu-desktop-installer/pull/1178
2. A subsequent crash after the previous one was fixed. https://github.com/canonical/ubuntu-desktop-installer/pull/1192

Both were fixed in the Ubuntu Desktop Installer repository, and this brings the latest fixes.


EduardGomezEscandell (approved):
LGTM 👍

PR #298, opened 2022-10-18
[Hotfix] Fixed e2e test expectedPolicy

EduardGomezEscandell:
This fixes an error that snuck in in #292


CarlosNihelton (approved):
LGTL :tada:

PR #301, opened 2022-10-19
Testing Psexec -h option for elevated token

toktamis:
-h option should avoid UAC issues. It is testing to see if UAC is the root cause for the access denied


toktamis:
It seems the account running psexec needs to run with more privilege. Need to check the Azure account.

PR #311, opened 2022-10-24
Addresses loss of comm between splash and launcher

CarlosNihelton:
The wait introduced in https://github.com/ubuntu/WSL/pull/278 broke the communication between the splash app (the slide show only) and launcher. The splash communicates its HWND on creation, through a named pipe created by the launcher. The wait caused the splash window to try to communicate before the named pipe was created. That wait is only useful if the child is a console app (the case for E2E). Thus, the appConfig check makes sense.

In the certainty that we will soon backport the OOBE on Windows to Jammy and thus deprecate the slide-show-only splash application, a more robust fix would be an overkill.


EduardGomezEscandell (approved):
LGTM 👍

PR #316, opened 2022-10-25
Fix CI failure due prefill file not found

CarlosNihelton:
In https://github.com/ubuntu/WSL/pull/297#issuecomment-1289899832 it was mentioned that CI started to present failures very often due launcher failing to write the prefill file.

This PR aims to address that condition by running a command through WSL API to make sure the distro is awakened before attempting to write the prefill file. Two successful runs in a row ( [this](https://github.com/ubuntu/WSL/actions/runs/3324081695) and [and this one](https://github.com/ubuntu/WSL/actions/runs/3324295570)  ) seem to confirm the suspicious is right.


EduardGomezEscandell (approved):
LGTM 👍

PR #319, opened 2022-11-01
Backporting systemd-sysusers.service fix to older distros

EduardGomezEscandell:
The service `systemd-sysusers` fails at startup when trying to load credentials. To fix this, we overrode its conf file in `/etc/systemd/system/sytemd-sysusers.d/override.conf` in UbuntuPreview.

This patch enables this fix in all distros. Note that systemd is still only enabled by default on Preview and Dev.

The function `Systemd::Enable` has been renamed to `Systemd::Configure` to reflect the fact that it permorms this action even when systemd is not enabled.


CarlosNihelton (approved):
LGTM :+1:

PR #322, opened 2022-11-08
Encapsulating VM setup in Azure in a Github action

EduardGomezEscandell:
End goal is for other projects to be able to use this action for their CI.


.github/workflows/e2e.yaml
```diff
@@ -2,44 +2,42 @@ name: End-to-End Tests
...
+  start-vm:
     strategy:
       max-parallel: 1
     runs-on: ubuntu-latest
     steps:
-      - uses: azure/login@v1
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Start Azure VM
+        uses: ubuntu/WSL/.github/actions/setup_vm/@ci-encapsulate-azure-wsl-309
```
EduardGomezEscandell, line 16: This change will have to be made before merging:
```suggestion
        uses: ubuntu/WSL/.github/actions/setup_vm/@main
```

I didn't manage to make these track the current branch tho, so I hard-coded them to test the CI.

didrocks: When tracking main, do you need `@main`? I didn’t need this [in this example](https://github.com/ubuntu/sync-issues-github-jira/blob/main/.github/workflows/example.yaml#L10). Note though that the event is different, being "gollum"

EduardGomezEscandell: Good point, let me try it.

.github/workflows/e2e.yaml
```diff
@@ -2,44 +2,42 @@ name: End-to-End Tests
 on:
   workflow_dispatch:
   pull_request:
-concurrency: build-wsl
+concurrency: e2e
```
didrocks, line 5: Shouldn’t that be "end-to-end-tests"?

EduardGomezEscandell: You can actually use any string here. From the docs:
> Use concurrency to ensure that only a single job or workflow using the same concurrency group will run at a time. A concurrency group can be any string or expression.
[...]
When a concurrent job or workflow is queued, if another job or workflow using the same concurrency group in the repository is in progress, the queued job or workflow will be pending.

I still changed it to your sugestion because I like that name better 😅.

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+  az_vm_password:
+    description: Password for "xfreerdp /p:az_vm_password"
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: rdp
+      shell: bash
+      run: |
```
didrocks, line 20: You should probably set -eu here, to ensure we stop as soon as something is failing and we don’t catch it.

Ditto for the other multi-lines scripts.

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+        set -o pipefail
+        sudo apt-get install -y xvfb freerdp2-x11
```
didrocks, line 22: You need to set that it’s non interactive to avoid debconf prompts (`DEBIAN_FRONTEND=noninteractive`)

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+        # detected empirically and confirmed here: https://github.com/FreeRDP/FreeRDP/blob/68ad8d5a1c8d52828095c2293acea1c76fa7fcb5/client/X11/xfreerdp.h#L350
+        vmOfflineErrorCode=141
```
didrocks, line 24: :mage:

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+  using: 'composite'
+  steps:
+    - name: rdp
+      shell: bash
+      run: |
+        set -o pipefail
+        sudo apt-get install -y xvfb freerdp2-x11
+        # detected empirically and confirmed here: https://github.com/FreeRDP/FreeRDP/blob/68ad8d5a1c8d52828095c2293acea1c76fa7fcb5/client/X11/xfreerdp.h#L350
+        vmOfflineErrorCode=141 
+        lastExitCode=$vmOfflineErrorCode
```
didrocks, line 25: well, here we know that it won’t happen, but can you try protecting every variables with:
- for non string, only `${}`
- for strings, protect against empty value with `"${}"`

(Same, not reperaing, but this goes throught the code)

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+        maxAttemptsCount=10;
+        while [[ $lastExitCode == $vmOfflineErrorCode && $maxAttemptsCount != 0 ]]; do
+          sleep 20 # seconds
+          xvfb-run -a xfreerdp /v:${{ inputs.az_vm_hostname }} /u:${{ inputs.az_vm_username }} /p:'${{ inputs.az_vm_password }}' \
+            /cert:tofu | grep -v ${{ inputs.az_vm_hostname }} || lastExitCode=$?
+          ((--maxAttemptsCount))
+        done
+
+branding:
```
didrocks, line 34: I didn’t find any icon matching this in https://github.com/ubuntu/WSL/actions/runs/3486268478/jobs/5832591830, where is it displayed? (curious)

.github/actions/session_nonzero/action.yml
```diff
@@ -0,0 +1,49 @@
+name: Azure Nonzero session execution
```
didrocks, line 1: I would rename the action to get straight to the point and not bother the "user" of the action with session 0 or none 0 (this could be in the description still)

.github/actions/session_nonzero/action.yml
```diff
@@ -0,0 +1,49 @@
...
+description: 'Execute a script as session >= 2.'
+
+inputs:
+  exec:
+    description: Script to execute
+    required: true
+  max-attempts:
+    description: Maximum number of attempts to acquire session > 2
```
didrocks, line 9: same here (and below, we should IMHO avoid the verbiage around session > 2). It’s just the maximum number of  connection attempts.

.github/actions/session_nonzero/action.yml
```diff
@@ -0,0 +1,49 @@
...
+    required: false
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Get session >= 2, and execute
+      shell: powershell
+      run: |
+        $higherSession=0
+        $count = 0
```
didrocks, line 20: nitpick hat on: mismatch in assignement style (see line above and this one)

.github/actions/session_nonzero/action.yml
```diff
@@ -0,0 +1,49 @@
...
+          $sessions=$(C:\SysinternalsSuite\logonsessions.exe -nobanner | Select-String Session: | Sort-Object -Descending | Get-Unique)
+          If ( $sessions -ne $null -and $sessions.length -gt 0 ) {
+            $higherSession=$( $sessions )[0].ToString()[-1]
+          }
+        }
+
+        c:\sysinternalssuite\psexec.exe -nobanner -accepteula -w $PWD.Path -i $higherSession powershell.exe -Command @'
+        &{
+          ${{ inputs.exec }}
+        } 3>&1 2>&1 > .\test.log
```
didrocks, line 38: There is nothing equivalent of tee, we can’t get the output directly here? (I’m afraid of jobs that takes multiple minutes and no feedback on the action meanwhile…)

.github/workflows/e2e.yaml
```diff
@@ -68,58 +66,48 @@ jobs:
...
+      - name: Run tests
+        uses: ./.github/actions/session_nonzero/
+        with:
+          exec: |
+            $env:WSL_UTF8=1
+            $env:LAUNCHER_REPO_ROOT=$PWD 
+            wsl.exe --version
+            cd .\e2e
+            go test .\test-runner -timeout 15m -run TestBasicSetup --distro-name ${{ env.distroName }} --launcher-name ${{ env.launcher}}
+            If ($LastExitCode) {
```
didrocks, line 90: Do you mind commenting on that for a non powershell expert? :p (it doesn’t exit with the last command value?)

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+        sudo apt-get install -y xvfb freerdp2-x11
+        # detected empirically and confirmed here: https://github.com/FreeRDP/FreeRDP/blob/68ad8d5a1c8d52828095c2293acea1c76fa7fcb5/client/X11/xfreerdp.h#L350
+        vmOfflineErrorCode=141 
+        lastExitCode=$vmOfflineErrorCode
+        maxAttemptsCount=10;
+        while [[ $lastExitCode == $vmOfflineErrorCode && $maxAttemptsCount != 0 ]]; do
+          sleep 20 # seconds
+          xvfb-run -a xfreerdp /v:${{ inputs.az_vm_hostname }} /u:${{ inputs.az_vm_username }} /p:'${{ inputs.az_vm_password }}' \
+            /cert:tofu | grep -v ${{ inputs.az_vm_hostname }} || lastExitCode=$?
+          ((--maxAttemptsCount))
```
didrocks, line 31: I’m unclear how this job stop btw once session_non_zero action is over?

.github/workflows/e2e.yaml
```diff
@@ -2,44 +2,42 @@ name: End-to-End Tests
...
+          az_vm_username: ${{ secrets.AZURE_VM_UN }}
+          az_vm_password: ${{ secrets.AZURE_VM_UP }}
 
-  e2e:
+  end-to-end-tests:
     strategy:
       max-parallel: 1
-    needs: start_vm
+    needs: start-vm
     runs-on: self-hosted
```
didrocks, line 41: You lost me a little bit here:
1. this is not in the description of the PR
2. even with this, how does this job knows on which machine to run this? This is supposively the VM that we started but I don’t find any resource group or anything that clarify this… So, unsure to understand really (we can hop on a HO, maybe that would be clearer, but I think the documentation should help too)

.github/workflows/e2e.yaml
```diff
@@ -2,44 +2,42 @@ name: End-to-End Tests
...
   workflow_dispatch:
   pull_request:
-concurrency: build-wsl
+concurrency: end-to-end-tests
+
 jobs:
-  start_vm:
+  start-vm:
     strategy:
       max-parallel: 1
```
didrocks, line 10: I really think we should eliminate this on the long run (let’s not block this PR on this) by:
- have a lock in the composite action itself to avoid multiple runs
- have the start VM idemnpotent
- maybe have a idling system on the VM to stop the machine? Then, we no longer rely on the last job to run.

We can then maybe further work in the composite action to eliminate the need for start-vm, have the first connect job doing it, and maybe make the second job depending on a particular step on the first one (this used to be possible in the early days of github action, but we need to see if this is still the case).

didrocks (changes requested):
Excellent work! This is already some great improvements over the first pass.

I have some questions (maybe they can be addressed in an easier way in a HO tomorrow) and some changes that can be made to make the script part more robust.

I think there are some opportunities for future larger enhancements, but let’s not block on that for now.

Also, a last comment I couldn’t post anywhere as its on the top description:  we don’t really need to checkout the code in the jobs apart from `insert-job-here` where the tests actually run, correct? (as the actions or github actions, treated as not needing to be local).

.github/actions/rdp/action.yaml
EduardGomezEscandell: Yeah I was very happy chosing icons and whatnot, but aparently they don't show up here 😞. They only appear in the Github Marketplace, if you chose to publish your action there.

.github/actions/session_nonzero/action.yml
EduardGomezEscandell: Apparently there is a `Tee-Object`. I'll try it out.

EduardGomezEscandell: I get the same result as I used to, for some reason nothing is printed to stdout. I'll have to investigate a bit more.

.github/actions/rdp/action.yaml
EduardGomezEscandell: It has its rug pulled under its feet by the `stop-vm` action, which shuts down the machine alltogether.

.github/workflows/e2e.yaml
EduardGomezEscandell: Agreed on all fronts. The Action that 3rd parties will want to write will be similar to the one in `wsl-exec-example.yaml` which is much simpler than the one in `e2e.yaml`.

.github/actions/session_nonzero/action.yml
EduardGomezEscandell: It fails because it is using RDP, so stdout/stderr are printed on the remote. I've quickly looked around the web and there seem to be ways of redirecting them from the remote to the local environment, but I'd leave that to the next PR

didrocks: interesting, I’m unsure to really understand why (for me, rdp was only for opening the session, but then, when you execute, your in another seat from the same session and get your output there), but that’s probably wrong :) anyway :+1: on addressing it in the next PR.

.github/workflows/wsl-exec.yaml
```diff
@@ -0,0 +1,184 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Vet input
+        env:
+          EXEC_TO_VET: ${{ inputs.exec }}
+        shell: powershell
+        run: |
+          If(${EXEC_TO_VET} -match "\n'@" -or ${EXEC_TO_VET} -match "^'@") {
+            Write-Error "Your script contains '@ at the begining of a line. This is not allowed because it would escape the Winows' quotes when invoking it."
```
EduardGomezEscandell, line 97:
```suggestion
            Write-Error "Your script contains '@ at the begining of a line. This is not allowed because it would escape the Windows' quotes when invoking it."
```

.github/workflows/e2e.yaml
EduardGomezEscandell: Scripts do indeed not exit with the last exit code. See this example:

```powershell
PS > cat .\deleteme.ps1
python.exe -c "exit(42)"
PS > powershell.exe -ExecutionPolicy Bypass -File .\deleteme.ps1
PS > echo $LASTEXITCODE
0
```

EduardGomezEscandell: Hving said this, those three lines of code can be (and now have been) simplified.

didrocks: My point was: do we really need this exit then? if `go test` returns 1, the script itself won’t return 1 (returning last execution status, as bash does)?

.github/actions/rdp/action.yaml
```diff
@@ -0,0 +1,69 @@
...
+runs:
+  using: 'composite'
+  steps:
+    - name: rdp
+      shell: bash
+      env:
+        DEBIAN_FRONTEND: noninteractive
+        MAX_CONNECTION_ATTEMPTS: "10"
+      run: |
+        set -eo pipefail
```
didrocks, line 24: Hum, you marked the previous comment as resolved (I don’t know, we tend to let the reviewer checking/mark things are resolved), I don’t see a -u here?

.github/workflows/wsl-exec.yaml
```diff
@@ -0,0 +1,184 @@
...
+        description: Pasword to the secret certificate
+        required: true
+    inputs:
+      exec:
+        description: Script to run on WSL
+        required: true
+        type: string
+      repo:
+        description: The url to clone the repo from
+        required: true
```
didrocks, line 33: Do we necessarily needs a repo for executing a script (as it can be an inline exec script)?

EduardGomezEscandell:
> the CI failed in case you didn't notice due to the RCP connection, mind having a look if this is due to not even retries or a more real issue?

It's a real issue, I'm working on it but it's tough nut to crack😅

.github/workflows/wsl-exec.yaml
EduardGomezEscandell: Fair enough. I made it optional so that cloning the repo can be skipped.

It's still useful because most CI runs will need to clone the repo in order to run tests on it.

EduardGomezEscandell:
Force pushed:
- To rebase to main (since we merged with #327 which is related to this one)
- Clean up the commit history because it was too messy.

.github/workflows/wsl-exec.yaml
```diff
@@ -0,0 +1,196 @@
...
+              git checkout ${{ inputs.checkout }}
+            '@
+            } else {
+              Write-Output "Repo not provided, skipping"
+               & ${{ env.LAUNCHER }} run mkdir action_working_dir
+            }
+            Write-Output '::endgroup::'
+
+            Write-Output '::group::Storing script into WSL'
+            $tmpdir="\\wsl.localhost\${env:distro}\tmp\"
```
CarlosNihelton, line 154: Executing from `/tmp` can easily become a security concern, specially because it's easy to conflict with files generated by other programs. I'd recommend create a temporary directory and file with `mktemp` inside Linux and use that instead of `/tmp/script`.

EduardGomezEscandell: Makes sense, however `mktemp` won't do it because it does not create real directories, or at least Windows thinks they are files rather than directories.

Since the WSL instance is volatile anyways, I can just use some random long name for the folder.

EduardGomezEscandell: `/tmp/github_${{ github.run_id }}` should be good enough

.github/workflows/wsl-exec.yaml
EduardGomezEscandell: Reverted this for simplicity. Repository is deduced automatically and as such there is no need for these inputs.

.github/workflows/wsl-exec-example.yaml
```diff
@@ -0,0 +1,28 @@
...
+    uses: ./.github/workflows/wsl-exec.yaml
+    secrets:
+      az_creds: ${{ secrets.AZURE_VM_CREDS }}
+      az_vm_hostname: ${{ secrets.AZURE_VM_AUTHORITY }}
+      az_vm_username: ${{ secrets.AZURE_VM_UN }}
+      az_vm_password: ${{ secrets.AZURE_VM_UP }}
+    with:
+      az_name: wsl-ci
+      az_resource_group: wsl
+      exec: |
```
CarlosNihelton, line 17: This looks really nice!

.github/actions/rdp/action.yaml
CarlosNihelton: DEBIAN_FRONTEND=noninteractive still missing.

CarlosNihelton (changes requested):
I think we are almost done. Just finish the `DEBIAN_FRONTEND=noninteractive` pointed out by Didier and we are good to go.

EduardGomezEscandell:
> I think we are almost done. Just finish the DEBIAN_FRONTEND=noninteractive pointed out by Didier and we are good to go.

That's done 😃 

The only usage of apt is here: https://github.com/ubuntu/WSL/blob/b275501fd76fcf8a8dffc4c5835b86063255890b/.github/actions/rdp/action.yaml#L20-L30
and the environment variable is set via github actions' `env`.

CarlosNihelton (approved):
LGTM :+1:

PR #324, opened 2022-11-14
Fixed memory leak in WslGetDistributionConfiguration

EduardGomezEscandell:
As per the documentation in https://learn.microsoft.com/en-us/windows/win32/api/wslapi/nf-wslapi-wslgetdistributionconfiguration#remarks
> The caller is responsible for freeing each string in pDefaultEnvironmentVariablesArray (and the array itself) via CoTaskMemFree.

Emphasis on "and the array itself".


EduardGomezEscandell (commented):
Testing which change breaks the tests

didrocks (approved):
CoTaskMemApprove(); !

The joy of C…

PR #325, opened 2022-11-24
Avoid triggering Refresh releases and assets

EduardGomezEscandell:
Refresh releases and assets is an automated action that looks for the string "UbuntuDev.WslID.Dev". If this string is found, then multiple copies of the file are generated in which this string is replaced with "Ubuntu-18.08", "Ubuntu-20.04", etc. To avoid triggering this, I changed the code so it does not contain string "UbuntuDev.WslID.Dev".

You can see an undesired trigger here: https://github.com/ubuntu/WSL/pull/290


EduardGomezEscandell:
Force-pushed to rebase to main so that #326 would take effect in the CI run

PR #326, opened 2022-11-24
Run update before installing in CI

didrocks:
We need to run apt update before running apt install in CI.

This should be picked in the new test workflow branch (with the `DEBIAN_FRONTEND=noninteractive` change too that I ignored here to avoid mix of concerns)


EduardGomezEscandell (approved):
🌙🦞

PR #327, opened 2022-11-24
[WSL-315] Refactoring end-to-end tests

EduardGomezEscandell:
This refactor aims to make the tests more idiomatic.

### Changes

These changes are so interlinked that creating a sequence of commits is a fool's errand. Instead, the refactor is one big commit:
- Changed some filenames
- Removed Tester struct
- Switched to using `require` package
- Created `test_cases.go`with reusable assertion sets that can be run as parallel subtests.
- Created test for `launcher.exe install --root --ui=none`.

The change to the CI can be isolated, I suppose:
- Modified CI to point to new files.


EduardGomezEscandell:
Here is some work left to do, but they all imply an expansion of functionality beyond the scope of this refactor. They should be solved in a follow-up PR.
- In TestBasicSetupFallbackUI: Try to conrtol stdin/stdout to avoid the need for `--root` installation. https://github.com/ubuntu/WSL/blob/dc8586a4cbfc4aea32faf061e94dca4dec6b3af0/e2e/test-runner/basic_setup_test.go#L43-L44
- In TestDefaultExperience: Find out how to control this remotely, or at least read stdout by redirecting it to a file. Also, analyse why this test fails on Azure but not on a regular machine. https://github.com/ubuntu/WSL/blob/dc8586a4cbfc4aea32faf061e94dca4dec6b3af0/e2e/test-runner/default_experience_test.go#L23-L25

e2e/test-runner/test_cases.go
```diff
@@ -140,3 +176,20 @@ func testUpgradePolicyIdempotent(t *testing.T) {
 
 	require.Equal(t, string(wantsDate), string(gotDate), "Launcher is modifying release upgrade every boot")
 }
+
+// systemdExpected returns true if systemd is expected to be enabled by default on this distro
+func systemdExpected() bool {
```
didrocks: Nitpick: I would even make it a sentence `systemdIsExpected()`

e2e/test-runner/default_experience_test.go
```diff
@@ -17,6 +17,17 @@ func TestDefaultExperience(t *testing.T) {
 	t.Skip("Skipped: fails in Azure") // TODO: Fix
 	wslSetup(t)
 
+	var installCompleted bool
+	t.Cleanup(func() {
+		if installCompleted {
```
didrocks: I think that reads better in one line: `if installCompleted && !t.Failed()`.

e2e/test-runner/test_utils.go
```diff
@@ -30,15 +32,8 @@ func wslSetup(t *testing.T) {
...
+	removeOldLogs(t)
 
-	rootDir := os.Getenv(constants.LauncherRepoEnvVar)
-	clientLogFullPath := filepath.Join(rootDir, clientLogPath)
-	err := os.Remove(clientLogFullPath)
-	if err != nil {
-		require.ErrorIs(t, err, os.ErrNotExist, "Failed to remove old install log at %q.", clientLogFullPath)
-	}
-
-	// Attempts to unregister the instance.
```
didrocks: Was the comment removal intended?

e2e/test-runner/test_utils.go
```diff
@@ -49,36 +44,52 @@ func wslSetup(t *testing.T) {
...
-		out, err = wslCommand(ctx, "cat", ServerLogPath).CombinedOutput()
-		t.Logf("%s", out)
-		if err != nil {
-			t.Logf("Failed to retrieve server debug log: %v", err)
-		}
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	out, err := wslCommand(ctx, "cat", subiquityAnswerFile).CombinedOutput()
+	if err != nil {
+		t.Logf("Failed to retrieve subiquity answer file: %v\n%s", err, out)
```
didrocks: Don’t print out if there is an error. The first argument content validity is not ensure in generally in Go when a function returns an error

e2e/test-runner/test_utils.go
```diff
@@ -49,36 +44,52 @@ func wslSetup(t *testing.T) {
...
-			t.Logf("Failed to retrieve client debug log: %v", err)
-		}
-	})
+	ctx, cancel = context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+	out, err = wslCommand(ctx, "cat", ServerLogPath).CombinedOutput()
+	if err != nil {
+		t.Logf("Failed to retrieve server debug log: %v\n%s", err, out)
+	} else {
+		s = s + fmt.Sprintf("\n==== Server Debug Log ====\n%s\n", out)
```
didrocks: ditto

e2e/test-runner/test_utils.go
```diff
@@ -49,36 +44,52 @@ func wslSetup(t *testing.T) {
...
+	}
+
+	out, err = ioutil.ReadFile(clientLogFullPath)
+	if err != nil {
+		t.Logf("Failed to retrieve client debug log: %v\n%s", err, out)
+	} else {
+		s = s + fmt.Sprintf("\n==== Client Debug Log ====\n%s\n", out)
```
didrocks: ditto

e2e/test-runner/test_cases.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // testUserNotRoot ensures the default user is not root.
-func testUserNotRoot(t *testing.T) {
+func testUserNotRoot(t *testing.T) { //nolint: thelper
```
didrocks: I would add an additional comment like ("This is a test"). Same below, but I won’t repeat it :)

e2e/test-runner/test_cases.go
```diff
@@ -85,13 +85,17 @@ func testSystemdEnabled(t *testing.T) {
...
+func testSystemdUnits(t *testing.T) { //nolint: thelper
 	if !systemdExpected() {
 		// Enable systemd in config file
-		wslCommandAsUser(context.Background(), "root", "bash", "-c", `printf '\n[boot]\nsystemd=true\n' >> /etc/wsl.conf`)
+		out, err := wslCommandAsUser(context.Background(), "root", "bash", "-c", `printf '\n[boot]\nsystemd=true\n' >> /etc/wsl.conf`).CombinedOutput()
+		require.NoError(t, err, "Failed to enable systemd: %s", out)
 		// Restore config file
-		defer wslCommandAsUser(context.Background(), "root", "bash", "-c", `head -n -2 "/etc/wsl.conf" > "/etc/wsl.conf"`)
+		t.Cleanup(func() {
+			out, err := wslCommandAsUser(context.Background(), "root", "bash", "-c", `head -n -2 "/etc/wsl.conf" > "/etc/wsl.conf"`).CombinedOutput()
```
didrocks: I’m not really confortable with that Cleanup approach. Nothing really ensure that the first two lines are what matters in the future.

Can we rather save the original vanilla file somewhere and restore it?

e2e/test-runner/test_utils.go
```diff
@@ -65,14 +65,14 @@ func subiquityLogs(t *testing.T) string {
...
 	defer cancel()
-	out, err = wslCommand(ctx, "cat", ServerLogPath).CombinedOutput()
+	out, err = wslCommand(ctx, "cat", serverLogPath).CombinedOutput()
 	if err != nil {
 		t.Logf("Failed to retrieve server debug log: %v\n%s", err, out)
 	} else {
 		s = s + fmt.Sprintf("\n==== Server Debug Log ====\n%s\n", out)
 	}
 
-	out, err = ioutil.ReadFile(clientLogFullPath)
```
didrocks: oh sorry, I didn’t spot you were using ioutil during our peer programming :) Be warned, as this change is like a couple of Go release ago, you will find a lot of examples still using ioutil, which is now deprecated (all functionalities have similar helpers in different other packages)

e2e/launchertester/test_cases.go
```diff
@@ -43,7 +43,7 @@ func testCorrectReleaseRootfs(t *testing.T) { //nolint: thelper
 		"Ubuntu-22.04":   "22.04",
 		"Ubuntu-20.04":   "20.04",
 		"Ubuntu-18.04":   "18.04",
-		"Ubuntu-Preview": "23.04",
+		"Ubuntu-Preview": "22.10",
```
didrocks: oh, why?

e2e/launchertester/test_cases.go
```diff
@@ -43,7 +43,7 @@ func testCorrectReleaseRootfs(t *testing.T) { //nolint: thelper
 		"Ubuntu-22.04":   "22.04",
 		"Ubuntu-20.04":   "20.04",
 		"Ubuntu-18.04":   "18.04",
-		"Ubuntu-Preview": "22.10",
+		"Ubuntu-Preview": "23.04",
```
didrocks, line 46: Ack :p

EduardGomezEscandell: We had a pull request race condition 😆

didrocks: Exactly :-)

e2e/test-runner/basic_setup_test.go
```diff
@@ -12,7 +12,7 @@ import (
 func TestBasicSetupGUI(t *testing.T) {
 	wslSetup(t)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Minute)
```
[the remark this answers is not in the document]
EduardGomezEscandell: Makes sense. I also added timeouts to other test cases to be more consistent.

e2e/test-runner/test_utils.go
EduardGomezEscandell: Yeah, I felt it was unnecessary given that the cleanup function is short and (from my perspective) quite clear. The comment added no extra info.

e2e/test-runner/test_utils.go
EduardGomezEscandell: I know it's bad in most other contexts, but isn't this quite standard for `Cmd.CombinedOutput`? What's the idiomatic way of reading the error message of a failed command?

e2e/test-runner/test_cases.go
EduardGomezEscandell: Yeah that makes much more sense. On it.

e2e/test-runner/test_utils.go
didrocks: Ah sorry, this is `CombinedOutput()`, so yeah, this is one exception when we don’t redirect stderr before. So, let’s keep it this way.

e2e/test-runner/test_utils.go
EduardGomezEscandell: Yep, the linter also complained so I fixed it later in this commit: 096600d8c41e3b697c1cf74f3159ff32b6b3fe5d

e2e/test-runner/default_experience_test.go
didrocks: oh right, that was || to get into the condition :p

didrocks (approved):
Looks excellent now! Thanks for the fixes. Really nice work :)

Issue #334, opened 2022-12-12
systemd-binfmt.service breaks interop


EduardGomezEscandell:
### cmd /c ver

Microsoft Windows [Versión 10.0.22623.1028]

### wsl --status

Distribución predeterminada: Ubuntu-22.04-test
Versión predeterminada: 2

### Did the problem occur during installation?

- [ ] Yes

### What happened?

systemd-binfmt.service remounts the `/proc/sys/fs/binfmt_misc/` directory, where `WSLInterop` is located and now overriden.

This directrory is shared across distros meaning that any distro can break all others' interop.

### What was expected?

Interop should not be broken by a service, much less across distros.

### Steps to reproduce

0. `wsl --shutdown`
1. Start any WSL distro that does not have systemd enabled.
2. Assert that `/proc/sys/fs/binfmt_misc/WSLInterop` exists.
3. Enable systemd and exit
4. `wsl --shutdown`
5. Start the distro
6. See that `/proc/sys/fs/binfmt_misc/` no longer contains `WSLInterop`.
7. See that `notepad.exe` fails.

the-moog:
It seems this issue is back in recent WSL. I'm experimenting with Wsl 1.3.11 with Ubuntu 23.04.  Just found the same issue.  It's probably time for a proper fix!
I'm not sure if the bug is in WSL runtime startup, Ubuntu distribution or WSL install (firstboot).
So raising it here as it's and older reference that still comes up in searches.
`ls -al /proc/sys/fs/binfmt_misc` shows only `python-3.1.1`  Of course `WSLInterop.conf` should be there too.
I know I posted a work-round elsewhere <del>(but I can't remember where)</del>  yes I can! 
workround: https://github.com/microsoft/WSL/issues/8952#issuecomment-1572193568

The suggestion here, that is disabling `systemd-binfmt.service`, is a bit of a blunt tool, e.g. it breaks other stuff, like Java.

Perhaps a proper long term fix would be to include a `WSLInterop.conf` file in `/etc/binfmt.d` through `systemd-firstboot`
or through the wsl overlays on `/usr/lib/binfmt.d` or in /run/binfmt.d (the latter may be best as it is not present on a stock Ubuntu 23 so likely will not break things).

NOTE this issues comes up a lot!!  Hence why I tried to find the earliest and link them.

- #8843
- #8952
- #9146
- #10172

PR #335, opened 2022-12-12
[DEENG-544] Disable systemd-binfmt.service

EduardGomezEscandell:
Disables systemd-binfmt.service to avoid breaking interop. This is a temporary solution, as we're still vulnerable to other distros breaking interop.

Fixes https://github.com/ubuntu/WSL/issues/334


CarlosNihelton (approved):
LGTM!

PR #337, opened 2022-12-15
[WSL-363] Allowing WSL Azure CI to be used outside the Ubuntu organization

EduardGomezEscandell:
## Raison d'etre
Reusable actions that are self-hosted cannot be used across different organizations. This means we need to re-think our approach.

## Basic structure
The only reason we had to use a reusable action was that we need the RDP and "Main" jobs to run concurrently. Hence, workflows will use three actions:
- `vm-setup`: This action starts the Azure VM and connects to the machine via RDP.
- Your main job goes here. This job runs concurrently to vm-setup so as to use the RDP session. You can use the "convenience actions" explained later.
- `vm-stop` stops the virtual machine. It runs after your main job, and concurrently to vm-setup, thereby killing the RDP session.

## Convenience actions
These actions have been created because I believe they have grat reusability potential. They are all prefixed with "wsl-" and I would continue this pattern should we develop any new ones.
- `wsl-install` installs or updates WSL and a distro of your choice (if any).
- `wsl-checkout` clones your repo into WSL.
- `wsl-bash` runs an arbitrary bash script within WSL (circumventing session 0 issues by piggybacking on the RDP session).
- `wsl-powershell` runs an arbitrary powershell script in Windows where WSL is allowed (circumventing session 0 issues by piggybacking on the RDP session).

## Customization
Users with heavier workloads may want to customize the start-stop cycle of their VMs. They can do that and still use the convenience actions which are our main contribution.

## Examples
See it in use here:
- Example usage: [wsl-example.yaml](https://github.com/ubuntu/WSL/blob/fix-paths-wsl-exec/.github/workflows/wsl-example.yaml)
- Real usage: https://github.com/EduardGomezEscandell/GoWSL/pull/10.

## Pending improvements
- `wsl-install` should be converted from a "composite" action to a "javascript" one. This would allow us to create a cleanup action that would uninstall and unregister the distro.
- `wsl-checkout` should be converted from a "composite" action to a "javascript" one. This would allow us to remove the repo in the cleanup.


EduardGomezEscandell:
There is one inconvenience:

When you use an action such as `actions/setup-go@v3`, Go is added to the path. However, if you connect via PsExec, this is not in the path for some reason (investigation pending).

Hence, you have to manually find the Go executable; or more simply have it installed in the machine already. This will be true for any setup-* action (think Go, Flutter, Python, etc.).

CarlosNihelton (changes requested):
I realize you are still working on this, but let's check those questions before proceeding.

.github/actions/vm-setup/action.yaml
```diff
@@ -15,6 +24,13 @@ inputs:
 runs:
   using: 'composite'
   steps:
+    - uses: azure/login@v1
+      with:
+        creds: ${{ inputs.az_creds }}
+    - name: Start the Runner
+      shell: bash
+      run: |
+        az vm start --name ${{ inputs.az_name }} --resource-group ${{ inputs.az_resource_group }}
```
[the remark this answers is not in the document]
EduardGomezEscandell: Well the available types are composite, javasript or docker; so we cannot control what OS this runs on. It'll fail later probably, I should document it.

https://docs.github.com/en/actions/creating-actions/about-custom-actions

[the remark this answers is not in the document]
EduardGomezEscandell: I understand `apt-get` will break it.

EduardGomezEscandell:
> I realize you are still working on this, but let's check those questions before proceeding.
I'm working on ironing out the last few details, last changes will be very small so feel free to review.

.github/actions/vm-setup/action.yaml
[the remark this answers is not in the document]
CarlosNihelton: Exactly. I believe a clear documentation should make it.

EduardGomezEscandell:
Example test run: https://github.com/EduardGomezEscandell/GoWSL/actions/runs/3892827764/jobs/6644724193

EduardGomezEscandell:
Example CI run:
https://github.com/EduardGomezEscandell/GoWSL/actions/runs/3894111684/jobs/6647673914
This is an expected failure.

For an example success, see either of these:
- https://github.com/ubuntu/WSL/actions/runs/3894059716/jobs/6647613467

.github/actions/vm-setup/action.yaml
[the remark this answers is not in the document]
EduardGomezEscandell: Alright, I added it to the readme.

EduardGomezEscandell:
Sometimes `Example action using WSL` will be auto-cancelled if  `End-to-End Tests` is already running because they share the same concurrency group. There is no way to wait instead of cancel, and it seems to be a race condition in our case.

Info on why it cancels: https://stackoverflow.com/a/74994502

.github/actions/wsl-install/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+              "Ubuntu"         { $storeId="9PDXGNCFSCZV" ; $launcher="ubuntu.exe"         ;  $appxname="Ubuntu"          }
+              "Ubuntu-Preview" { $storeId="9P7BDVKVNXZ6" ; $launcher="ubuntupreview.exe"  ;  $appxname="UbuntuPreview"   }
+              "Ubuntu-22.04"   { $storeId="9PN20MSR04DW" ; $launcher="ubuntu2204.exe"     ;  $appxname="Ubuntu22.04LTS"  }
+              "Ubuntu-20.04"   { $storeId="9MTTCL66CPXJ" ; $launcher="ubuntu2004.exe"     ;  $appxname="Ubuntu20.04LTS"  }
+              "Ubuntu-18.04"   { $storeId="9PNKSF5ZN4SW" ; $launcher="ubuntu1804.exe"     ;  $appxname="Ubuntu18.04LTS"  }
+              default          { Write-Output "Error: Unknown distro ${distro}" ; Exit(1) }
+          }
+
+          # Unregister and uninstall pre-existing distro
+          Remove-AppxPackage "$(Get-AppxPackage | Where-Object Name -like "CanonicalGroupLimited.${appxname}")" 3>&1 2>&1 | Out-Null
```
didrocks, line 38: You are using like here, IIRC, you told that it was replaced with "foo*", so it means that if we set the name to `Ubuntu`, can this may be transformed into `-like "CanonicalGroupLimited.Ubuntu*"` and thus match any other distro?

.github/actions/wsl-powershell/action.yaml
```diff
@@ -0,0 +1,100 @@
...
+        if ( $(Test-Path -Path ".\output.log") ) {
+          Get-Content ".\output.log"
+        } else {
+          Write-Output "No log"
+        }
+
+        Write-Output "----------Exit status-----------"
+        if ( ! $(Test-Path -Path ".\exitCode.log") ) {
+          Write-Output "No exit status reported, assuming failure"
+          Exit(-1)
```
didrocks, line 86: We can exit negative number in powershell? No side-effect?

.github/workflows/wsl-example.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+      - name: Install or update WSL
+        uses: ./.github/actions/wsl-install
+        with:
+          distro: Ubuntu
+
+      - name: Checkout your repo into WSL
+        uses: ./.github/actions/wsl-checkout
+        with:
+          distro: ubuntu
```
didrocks, line 55: So, we have some jobs with the same parameter but not the same meaning if I’m correct:
- in `wsl-install`, the `distro` parameter is the appx to install (Name), hence `Ubuntu`
- in the other steps, the `distro` parameter of the actions is the wsl instance that we want to run it. I think you are using `ubuntu` here, no capital, because it’s the name the distro is registered under? Or it doesn’t matter (and in this case we should use the same case so that we don’t loose the user).

didrocks (commented):
As usual, excellent job, really well documented to. Nice to see some CI like this we can reuse :) I have a couple of questions before the finale approval. The last one being the most important I guess.

.github/actions/wsl-install/action.yaml
EduardGomezEscandell: The opposite. It used to be `*distro-name*`, now it is an exact match. I could change it to 
```diff
- Where-Object Name -like "CanonicalGroupLimited.${appxname}"
+ Where-Object Name -value "CanonicalGroupLimited.${appxname}"
```
but the result is the same (-like and -value do the same thing if there are no asterisks).

.github/actions/wsl-powershell/action.yaml
EduardGomezEscandell: It underflows so it shows up as 0xffffffff

.github/workflows/wsl-example.yaml
EduardGomezEscandell: Distro always refers to the value that appears when you do `wsl --list --verbose`.

It's windows, so not case-sensitive. You're right that it is confusing, I'll change it to always be `Ubuntu` since that is what shows up when you run the aforementioned command.

.github/actions/wsl-install/action.yaml
didrocks: Ok, I was afraid of the unexpected regexp behaviour completing the end :)

CarlosNihelton (approved):
Well-done!
LGTM 👍🏽  once addressed the case in Ubuntu name.

didrocks (approved):
Excellent! Feel free to merge once the tests pass ofc :)

EduardGomezEscandell:
Yeah seems like the rootfs download is timing out. Will keep retrying for now.

We should soon implement some sort of caching, it takes too long anyways.

PR #338, opened 2022-12-16
Updates the splash submodule

CarlosNihelton:
To enable statically linking Flutter binaries against VC libs. To avoid the package dependency and better support Windows Server.


EduardGomezEscandell (approved):
👍 C'est bon

CarlosNihelton:
> 👍 C'est bon

You are preparing a French revolution for the next print, aren't you? Hahaha

PR #339, opened 2022-12-16
Updates UDI submodule

CarlosNihelton:
To statically link VC libs and remove the package dependency. To better support Windows Server.

There are a bunch of changes since last submodule update, but the interesting ones for us is the fact that the last screen is no longer shown.


EduardGomezEscandell (approved):
👍 é bom

PR #340, opened 2022-12-16
Removes the `PackageDependency` entry from manifest

CarlosNihelton:
To support Windows Server, we cannot rely on the store to deliver VClibs.
The launcher vcxproj is already configured the way we want. The Flutter binaries are now, since #338 and #339 .


EduardGomezEscandell (approved):
🧁 Great!

PR #344, opened 2023-01-12
Phasing out the hybrid OOBE

CarlosNihelton:
We had an additional complexity when it comes to the OOBE because UbuntuPreview had one architecture while the 22.04 LTS had another one.

This PR phases out the LTS version, in which we had a Windows application displaying a slide show while the setup wizard was a Linux app rendered through WSLg.

For that we:

1. Remove the extra conditional compilation that used to look for `WIN_OOBE_ENABLED` in `ApplicationStrategy.h`. Now the conditional is only whether the architecture supports the GUI or not.
2. Make the default `DistroLauncher-Appx/Directory.Build.props` file to build the OOBE instead of the splash. Achieved by copying the definition from the `meta/UbuntuPreview` override.
3. Updated the `ubuntu-desktop-installer` submodule to bring the recent changes that allow skipping some pages.

Additionally this removes  the `ubuntu-wsl-splash` submodule and the referred UbuntuPreview override, since it wouldn't override anything.

Some dead code in the C++ sources is left untouched to be handled in a next pull request.


EduardGomezEscandell (approved):
Always nice to have pull requests that add my removing ^^. LGTM 👍.

PR #346, opened 2023-01-13
Fixed internal imports in actions

EduardGomezEscandell:
Internal imports need to be absolute for external repositories to be able to use them.

Follow-up to #337.

### DIFF
--- a/.github/actions/wsl-bash/action.yaml
+++ b/.github/actions/wsl-bash/action.yaml
@@ -38,7 +38,7 @@ runs:
         if ( "${exit}" -eq "$False" ) { Exit(1) }
 
     - name: Run linux script
-      uses: ./.github/actions/wsl-powershell
+      uses: Ubuntu/WSL/.github/actions/wsl-powershell
       with:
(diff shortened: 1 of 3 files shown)


.github/actions/wsl-bash/action.yaml
```diff
@@ -38,7 +38,7 @@ runs:
         if ( "${exit}" -eq "$False" ) { Exit(1) }
 
     - name: Run linux script
-      uses: ./.github/actions/wsl-powershell
+      uses: Ubuntu/WSL/.github/actions/wsl-powershell@main
```
CarlosNihelton, line 41: Honest question: is this syntax case sensitive? I'd expect it to fail because the org name is capitalized.

EduardGomezEscandell: Apparently not, because the "example action using WSL" passed. I'll still fix it just to make it cleaner,

CarlosNihelton (approved):
LGTM 👍🏽

PR #348, opened 2023-01-23
Updates rootfs URL generation for Jammy

CarlosNihelton:
Previously only Kinetic was published on WSL pipeline. Jammy has also been ported.
This PR changes the URL generation in `wsl-builder/prepare-build` to acknowledge the new rootfs location.

That's the reason why the CI is failing to build the 22.04 LTS app.


EduardGomezEscandell (approved):
Also fixed here https://github.com/ubuntu/WSL/pull/345 (commit [d9c1d96](https://github.com/ubuntu/WSL/pull/345/commits/d9c1d96f86e34ff101c1e36fda9343e7dd4c865e)), but let's not wait.

PR #352, opened 2023-02-08
(fix) Cannot read output from previous job steps

CarlosNihelton:
Basic GitHub actions mistake. 🙈 

We are attempting to store the debug symbols of the latest published apps in the wiki.
For that we need to know whether this build lead to a new submission to the store: `steps.detect-upload-to-store.outputs.needs-upload`.

The problem is that we were attempting to read the output of a step that belongs to another job. That can only work if the job who owns that step makes the output available.

### DIFF
--- a/.github/workflows/build-wsl.yaml
+++ b/.github/workflows/build-wsl.yaml
@@ -89,6 +89,8 @@ jobs:
     env:
       buildInfoPath: 'wiki/build-info'
       workDir: 'C:/Temp/builddir'
+    outputs:
+      store-needs-upload: ${{ steps.detect-upload-to-store.outputs.needs-upload }}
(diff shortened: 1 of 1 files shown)


CarlosNihelton:
CI break seems unrelated, so I'm considering this ready for review.

EduardGomezEscandell (approved):
I see you too are a yaml programmer these days

EduardGomezEscandell:
I think CI error is a problem with winget, we may need to revisit the `wsl-install` action. Winget fails when no updates exist. Indeed something for a different PR.

CarlosNihelton:
> I see you too are a yaml programmer these days

I trust your authority in this topic 😄

PR #355, opened 2023-03-17
Having no updates to install should not break CI

CarlosNihelton:
WSL-449

That value is winget's exit code when no updates are available to install. That means we have the latest and greatest installed. No reason for immediate crashing.
That error code might mean a real failure, though. Thus as an extra precaution we launch wsl --version and check.

See

https://github.com/microsoft/winget-cli/blob/bfcee8edba2853f5a11e96dd7c7f6b412afef26a/src/AppInstallerSharedLib/Public/AppInstallerErrors.h#L58

And

https://github.com/microsoft/winget-cli/blob/bfcee8edba2853f5a11e96dd7c7f6b412afef26a/src/AppInstallerCLICore/Workflows/UpdateFlow.cpp#L67

### DIFF
--- a/.github/actions/wsl-install/action.yaml
+++ b/.github/actions/wsl-install/action.yaml
@@ -12,13 +12,11 @@ runs:
   steps:
     - name: Install WSL
       shell: powershell
+      continue-on-error: true
       run: |
         # Install or update WSL
         $env:WSL_UTF8=1
-        # Having no updates to install should not be an error.
-        $NoUpdateError=0x8A15002B
-        winget install --name 'Windows Subsystem for Linux' --accept-source-agreements --accept-package-agreements --silent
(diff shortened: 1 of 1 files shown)


CarlosNihelton:
I'm waiting on an updated rootfs that will allow CI to pass tests.

EduardGomezEscandell:
Remember that the rootfs on out Azure machine is cached and refreshed daily, so you may need to trigger a refresh manually https://github.com/ubuntu/WSL/actions/workflows/update-rootfs-cache.yaml once it's up on Ubuntu Cloud images in case it was uploaded after the scheduled refresh.

CarlosNihelton:
I rebased this on top of #361 because none of those two PRs alone would suffice to make the CI pass, but both together plus a new rootfs not yet available.

PR #361, opened 2023-04-14
Updates UDI due PR #1812

CarlosNihelton:
Once a new lunar image becomes available the end-to-end tests will still hang even though the installation will succeed. 

https://github.com/canonical/ubuntu-desktop-installer/pull/1812 removes two lines of test code to prevent such behavior.

This PR updates UDI submodule from latest main to incorporate that fix.


EduardGomezEscandell (approved):
Good stuff 👍

Interesting that the number of CPU is relevant. Idk why it would matter.

CarlosNihelton:
> Good stuff 👍
> 
> Interesting that the number of CPU is relevant. Idk why it would matter.

I guess that's a high parallelized operation. I only suspected that improvement would be needed when I ran the end-to-end test script many times while watching the resource consumption through the task manager. I could see memory was ok, but CPU was at 100% all the time and the test failing awkwardly.

PR #366, opened 2023-04-25
Removes the dependency on the UAP10 namespace

CarlosNihelton:
Deploying Ubuntu 22.04 to Windows Server 2019 fails due what seems to be an invalid `AppxManifest.xml`. The truth behind that message is that the namespace `http://schemas.microsoft.com/appx/manifest/uap/windows10/10` is not supported on that OS version.

We only used that namespace to pass command line parameters to the application.

This PR:
1. changes the CLI parsing to ensure the previously desired behavior is now matched when no CLI parameters are passed 
2. Removes the dependency on that namespace from the `MyDistro.appxmanifest` of all LTSs and the main Ubuntu app.
3. Preserves the above dependency for the Ubuntu Preview application.

The point 3 ensures that Ubuntu Preview invokes the OOBE by default. That behavior is being herein preserved under the assumption that the users willing to install Ubuntu WSL apps on Windows Server SKUs won't play with the unstable `Ubuntu Preview` app.

I did a manual sanity check over 22.04, the default Ubuntu, 18.04 and Preview to ensure the intended behavior remains. The gist of it is:

- Ubuntu Preview by default invokes the OOBE.
- The LTSs and Ubuntu default app by default go to the upstream setup. They go otherwise to OOBE if invoked with the `--ui` option.


EduardGomezEscandell (approved):
LGTM 👍

jibel:
currently testing a build of Preview fron this branch

jibel:
tested successfully Ubuntu-Preview with --ui=none

PR #369, opened 2023-04-26
Applies all relevant patches to the distro being installed

CarlosNihelton:
A top-level function called from `InstallDistribution()` (see `DistroLauncher.cpp`). Calling at this place we can ensure we have the necessary file system permissions to perform any config patch.

Calls patches in two steps: globals and release specific.

The patch collections are defined, but empty for now. Thus, for now this does nothing.

The collections will be populated as we add specific patches at next.

Nothing will change on this function or its caller as we add more patches.


EduardGomezEscandell (approved):
Full steam ahead 🛥️

PR #370, opened 2023-04-26
Removes `cloudimg-rootfs` label from all distros fstab's

CarlosNihelton:
A patching function is implemented and registered for all distros to remove all lines from `/etc/fstab` which start with `LABEL=cloudimg-rootfs`.

The existence of that line causes the `systemd-remount-fs.service` to fail. That's been observed on 20.04 but might happen to any distro.

Builds on top of the Patcher machinery presented in #368 .

Now it becomes easy to understand how that whole thing pays off (IMHO).

### DIFF
--- a/DistroLauncher/algorithms.h
+++ b/DistroLauncher/algorithms.h
@@ -130,3 +130,33 @@ template <typename T, typename... Args> void push_back_many(std::vector<T>& vec,
     vec.reserve(vec.size() + sizeof...(args));
     (vec.push_back(std::forward<Args>(args)), ...);
 }
+
+// Follows the semantics of std::getline(stream&, string&) while implemented for an iterator instead of the stream
+// itself, i.e. reads from [input] iterator until [delim] is found. Although it doesn't go to the string, it's
+// extracted from the underlying stream pointed by the [input] iterator. [str] is also erased before insertion.
+template <class CharT, class Traits, class Allocator>
+std::istreambuf_iterator<CharT, Traits>& getline(std::istreambuf_iterator<CharT, Traits>& input,
+                                                 std::basic_string<CharT, Traits, Allocator>& str, CharT delim = '\n')
+{
+    str.erase();
+    // missing std::copy_until - a copy_if that stops when the predicate evaluates to true :(
+    while (input != std::istreambuf_iterator<CharT, Traits>{}) {
+        char c = *input++;
+        if (c == delim) {
+            break;
+        }
+        str.append(1, c);
+    }
+    return input;
+    // We could, but let's not make it a template for now (at least until we find a need for reading another char type).
+    // We may find ourselves creating a std::istream just to call widen('\n') on it. Less then optimal.
+}
+
+// Returns a string view that starts on the first non-space element of [str] until its end.
+template <class CharT, class Traits>
+std::basic_string_view<CharT, Traits> left_trimmed(std::basic_string_view<CharT, Traits> str)
+{
+    auto nonSpaceStart = std::find_if_not(str.begin(), str.end(), [](auto elem) { return std::isspace(elem); });
+    auto trimmedSize = static_cast<std::size_t>(std::distance(nonSpaceStart, str.end()));
+    return std::basic_string_view<CharT, Traits>{&(*nonSpaceStart), trimmedSize};
+}
--- a/DistroLauncher/algorithms.h
+++ b/DistroLauncher/algorithms.h
@@ -134,8 +134,11 @@ template <typename T, typename... Args> void push_back_many(std::vector<T>& vec,
 // Follows the semantics of std::getline(stream&, string&) while implemented for an iterator instead of the stream
 // itself, i.e. reads from [input] iterator until [delim] is found. Although it doesn't go to the string, it's
 // extracted from the underlying stream pointed by the [input] iterator. [str] is also erased before insertion.
+// Iterator is taken by copy, as all algorithms accepting iterators are, but one must remember that the underlying
+// stream linked to the iterator will still have it's internal state changed as the iterator advances, i.e.
+// contents are going to be extracted.
 template <class CharT, class Traits, class Allocator>
-std::istreambuf_iterator<CharT, Traits>& getline(std::istreambuf_iterator<CharT, Traits>& input,
+std::istreambuf_iterator<CharT, Traits> getline(std::istreambuf_iterator<CharT, Traits> input,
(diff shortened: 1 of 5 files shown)


DistroLauncher-Tests/DistroLauncher-Tests/PatchConfigTests.cpp
```diff
@@ -172,4 +179,54 @@ options=metadata
...
+        std::istringstream input(slightlyChanged);
+        std::stringstream output;
+
+        // Apply
+        PatchingFunctions::RemoveCloudImgLabel(input, output);
+
+        // Assert
+        // the patch function should have preserved the other line, which is just a comment.
+        EXPECT_EQ(output.str(), sampleContents(SampleStrings::confComment));
+    }
```
EduardGomezEscandell, line 208: Missing tests:
- fstab has leading whitespace (since you implemented behaviour to consider this).
- fstab does not contain the fstab1804 line

CarlosNihelton:
```cpp
        std::string slightlyChanged{sampleContents(SampleStrings::confComment)};
        slightlyChanged.append("\n    ");
        slightlyChanged.append(sampleContents(SampleStrings::fstab1804));
```

That was causing the second line to start with spaces. But I'll clarify that test case.

DistroLauncher/algorithms.h
```diff
@@ -130,3 +130,33 @@ template <typename T, typename... Args> void push_back_many(std::vector<T>& vec,
     vec.reserve(vec.size() + sizeof...(args));
     (vec.push_back(std::forward<Args>(args)), ...);
 }
+
+// Follows the semantics of std::getline(stream&, string&) while implemented for an iterator instead of the stream
+// itself, i.e. reads from [input] iterator until [delim] is found. Although it doesn't go to the string, it's
+// extracted from the underlying stream pointed by the [input] iterator. [str] is also erased before insertion.
+template <class CharT, class Traits, class Allocator>
+std::istreambuf_iterator<CharT, Traits>& getline(std::istreambuf_iterator<CharT, Traits>& input,
+                                                 std::basic_string<CharT, Traits, Allocator>& str, CharT delim = '\n')
```
EduardGomezEscandell, line 139: Not convinced with the function signature. In general, iterators are passed by copy, and an updated copy is returned. So `it = getline(it, str)` is the typical form (see std::find, std::partition, std::merge, etc).

Alternatively we could sin a little by passing the iterator by reference and returning the string: `const std::string str = getline(it)`, to help with const-correctness. I would not go for this option because modifying an iterator argument is anti-idiomatic.

EduardGomezEscandell: Oh you're right, I didn't know getline took iterators. Then your implementation is the most idiomatic.

CarlosNihelton:
Rebased over latest main, but not without observing your recommendation about the iterator taken by value on `getline`.

CarlosNihelton:
Looks like we are failing to acquire a milestone date for the next release from Launchpad. Which is not surprising at this time.

DistroLauncher/algorithms.h
```diff
@@ -130,3 +130,36 @@ template <typename T, typename... Args> void push_back_many(std::vector<T>& vec,
...
+// extracted from the underlying stream pointed by the [input] iterator. [str] is also erased before insertion.
+// Iterator is taken by copy, as all algorithms accepting iterators are, but one must remember that the underlying
+// stream linked to the iterator will still have it's internal state changed as the iterator advances, i.e.
+// contents are going to be extracted.
+template <class CharT, class Traits, class Allocator>
+std::istreambuf_iterator<CharT, Traits> getline(std::istreambuf_iterator<CharT, Traits> input,
+                                                std::basic_string<CharT, Traits, Allocator>& str, CharT delim = '\n')
+{
+    str.erase();
+    // missing std::copy_until - a copy_if that stops when the predicate evaluates to true :(
```
EduardGomezEscandell, line 145: This would be the way if you want to be an algorithm fanboy 😆.
```c++
const auto end = std::find(input, istreambuf_iterator{}, delim);
std::copy(input, end, std::ostream_iterator(str));
return end;
```

DistroLauncher/algorithms.h
```diff
@@ -130,3 +130,36 @@ template <typename T, typename... Args> void push_back_many(std::vector<T>& vec,
...
+    return input;
+    // We could, but let's not make it a template for now (at least until we find a need for reading another char type).
+    // We may find ourselves creating a std::istream just to call widen('\n') on it. Less then optimal.
+}
+
+// Returns a string view that starts on the first non-space element of [str] until its end.
+template <class CharT, class Traits>
+std::basic_string_view<CharT, Traits> left_trimmed(std::basic_string_view<CharT, Traits> str)
+{
+    auto nonSpaceStart = std::find_if_not(str.begin(), str.end(), [](auto elem) { return std::isspace(elem); });
```
EduardGomezEscandell, line 162:
```suggestion
    auto nonSpaceStart = std::find_if_not(str.begin(), str.end(), std::isspace);
```
Pedantic stuff.

EduardGomezEscandell (approved):
Some bikeshedding comments, not blocking.

DistroLauncher/algorithms.h
CarlosNihelton: `iostream`s and their iterators are supposed to be single pass. So that would require remembering the stream position at start and `seek` back to it prior to calling std::copy (raw calls to the `istream` or the `istreambuf`), otherwise `input` and `end` might both end up in the end of the stream if `delim` is not found or `input` becomes one past `end`, if `delim` is found. Interleaving `tellg` and `seek` calls would fix the issue, but then we read the file twice 😞 : (unless compiler does some optimization to inline both algorithms and interleave them, which I don't doubt that may happen in optimized builds, but not in debug and it's hard to rely on that fact).

_"No raw loops"_ not always work.

DistroLauncher/algorithms.h
CarlosNihelton: That doesn't work because of the const locale parameter :)

```cpp
template <class _Elem>
bool(isspace)(_Elem _Ch, const locale& _Loc) {
    return _STD use_facet<ctype<_Elem>>(_Loc).is(ctype_base::space, _Ch);
}
```

EduardGomezEscandell: So close to being great 😞

PR #371, opened 2023-04-27
Fix  `systemd-modules-load.service` on 18.04 via `ConditionVirtualization` for units

CarlosNihelton:
Implements a patching function to write overrides to units preventing them to start on containers. Such as WSL distro instances.
That function is registered for `systemd-modules-load.service` on 18.04. Other releases already have that condition on the system unit file shipped with the distro rootfs.


EduardGomezEscandell (approved):
:guitar:

PR #372, opened 2023-04-27
Disables `multipathd.socket` on Ubuntu-20.04

CarlosNihelton:
By reusing `OverrideUnitVirtualizationContainer()` supplied in #371 but registering it to write an override for `multipathd.socket`.

When enabling systemd on Ubuntu 20.04, `multipathd.service` reports failure because of the socket. The service itself has the `ConditionVirtualization` on the unit file, but the socket doesn't. Something else might be trying to talk through that socket, causing systemd to try to start it and fail. Only 20.04 was observed with that behavior.

Before:
Captura de tela 2023-04-26 171350

After:
Captura de tela 2023-04-26 225809


EduardGomezEscandell (approved):
Short and sweet.

PR #373, opened 2023-04-28
Fixes for `FlutterGuardedError` seen in CI

CarlosNihelton:
This PR addresses further issues with CI end-to-end tests:

1. `FlutterGuardedError` spuriously appearing. Recently became really recurrent. Partially addressed in https://github.com/canonical/ubuntu-desktop-installer/pull/1887.
2. Ensures the submodules are recursively updated for the end-to-end tests. The other workflows where this might be relevant update the submodules via git CLI.

While investigating 1 I noticed leftovers not being updated/removed from the UDI submodule. I suppose those leftovers are somehow related with the higher frequency the CI failed the past three days.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -33,6 +33,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell (approved):
👍

PR #375, opened 2023-04-28
Deletes code related to controlling the splash application

CarlosNihelton:
Since #344 we no longer support the hybrid OOBE, which was partially composed of a Windows application (a.k.a. the splash application) and partially a Linux GUI presented through WSLg.

This PR aims to remove the most obvious functions and classes that only served the purpose of leveraging that model and now are completely unused.

Smaller code base means faster compilation, but also faster linting.

There will still be some other pieces of code that are currently in use as they are, but subject to a great deal of simplification once we realize that the complexity was due the hybrid model. I'm referring to the installer controller state machine. But that's a topic for a future pull request.


EduardGomezEscandell (approved):
Always nice to delete code :D

Otherwise at some point we need to have directories in this repo, at least for the code we add on top of the upstream.

PR #378, opened 2023-05-02
Fix CI crash due empty milestone.date_targeted

CarlosNihelton:
This change should prevent CI from crashing if the current development release doesn't have yet a milestone target date.

Example of a crash: https://github.com/ubuntu/WSL/actions/runs/4859862714/jobs/8663001776?pr=372


jibel (approved):
Reviewed together. Approved.

PR #384, opened 2023-05-09
Simplify CI by avoiding use of RDP

EduardGomezEscandell:
As proven in https://github.com/ubuntu/GoWSL/pull/28, we do not need a remote desktop connection to run the CI. We can set up our machine to automatically log on, and set the GitHub runner as a startup program. It'll run in session >= 2, meaning we can use WSL.

This PR shows the changes in the codebase, but there also need to be changes in our machine. Namely:
1. Set up autologon in the VM: https://learn.microsoft.com/en-us/troubleshoot/windows-server/user-profiles-and-logon/turn-on-automatic-logon
2. Add the GitHub runner as a startup program in the VM.

UDENG-287


CarlosNihelton (approved):
This looks great!

We earn so much more benefits from the simplification than downsides 🚀 

The autologon could represent a risk if the runner was allowed to run when not executing those workflows. But that's not our case. We have tight control over the runner lifecycle.
When the day arises, other projects targeting WSL will benefit a lot from this work. Backed by proper documentation, it will be easy to reuse this.

EduardGomezEscandell:
> Backed by proper documentation

Comming soon :D

PR #387, opened 2023-05-11
(fix): Pin Flutter Version in CI

CarlosNihelton:
Aligned with UDI, because some dependencies are broken with Flutter 3.10 (just released).
Also, CI expected Ubuntu-Preview to be Lunar, but we are now at Mantic.
Having two unrelated fixes together is the only way to see the CI checks all going green :)


EduardGomezEscandell:
Great! As long as we don't squash the commits we're good :D.

EduardGomezEscandell (approved):

PR #389, opened 2023-05-19
Bugfix: String view end iterator is dereferenced in left_trimmed algorithm

EduardGomezEscandell:


CarlosNihelton (approved):
Thanks for the fix! 

We know already where the CI is failing and how to approach the quintessential fix!

CarlosNihelton:
Forgot to recommend squashing it because I played with your branch a little bit while you were off :D

PR #390, opened 2023-05-23
Add debugging info display to the notify-isotracker github action.

sil2100:
We're trying to debug why recent builds did not result in isotracker notifications. This will give a bit more context at least.


didrocks (approved):
Thanks @sil2100!

PR #392, opened 2023-05-23
Gardening 🌱:  Improve WSL checkout on private repos

EduardGomezEscandell:
It's still not supported, but at least it wont hang forever.


didrocks (approved):
Indeed, this is better that way :)

PR #393, opened 2023-05-23
Fix miscelaneous CI issues

EduardGomezEscandell:
A few issues have arised with our CI since #384 was merged. These are solved here.

### DIFF
--- a/.github/actions/wsl-install/action.yaml
+++ b/.github/actions/wsl-install/action.yaml
@@ -13,16 +13,12 @@ runs:
     - name: Install WSL
       shell: powershell
       continue-on-error: true
-      env:
-        WSL_UTF8: "1"
       run: |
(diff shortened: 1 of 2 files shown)


CarlosNihelton:
The workflow that is failing now is affected by an incompatibility introduced in one transitive dependency (logging_appenders). The fix requires an upgrade in the UDI submodule and Flutter SDK as well, so I'd rather have that done in a separate PR, since it's a little more evolving.

Details in:

https://github.com/canonical/ubuntu-desktop-installer/commit/62a5ead0c592cd77a2876e9634b9fa73dc9c94e6

https://github.com/canonical/ubuntu-flutter-plugins/commit/6d74e3d5699e5b31b1fd56983b71ff01ff2ac625

CarlosNihelton (approved):
LGTM 👍🏽 This is heading to a very nice direction!

PR #394, opened 2023-05-23
fix: Disable SSH unit on 20.04 via `systemctl disable`

CarlosNihelton:
After a team discussion we concluded that the way we closed #382 was inappropriate (thus this PR reverts it) and we should disable the SSH `systemd` unit via `systemctl disable`.

That cannot be made into the existing patching feature.

So I took the simplest approach of just running the command explicitly, but I placed the code where it's easily amendable to a more involved solution, if we start to mix and match different releases and `systemd` units to disable and require something similar to the patching feature into selecting the units per release and forming the `systemctl`command line accordingly. For now that would be an overkill since we only have one `systemd` unit and one release to match.

---

The result of a sideload installation of 20.04 after this looks like the following:

```
Installing, this may take a few minutes...
Please create a default UNIX user account. The username does not need to match your Windows username.
For more information visit: https://aka.ms/wslusers
Enter new UNIX username: u
New password:
Retype new password:
passwd: password updated successfully
The operation completed successfully.
Installation successful!
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

Welcome to Ubuntu 20.04.6 LTS (GNU/Linux 5.15.90.1-microsoft-standard-WSL2 x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

  System information as of Tue May 23 15:19:55 -03 2023

  System load:  2.97                Processes:             69
  Usage of /:   0.1% of 1006.85GB   Users logged in:       0
  Memory usage: 3%                  IPv4 address for eth0: 172.25.228.251
  Swap usage:   0%

Expanded Security Maintenance for Applications is not enabled.

0 updates can be applied immediately.

Enable ESM Apps to receive additional future security updates.
See https://ubuntu.com/esm or run: sudo pro status


The list of available updates is more than a week old.
To check for new updates run: sudo apt update


This message is shown once a day. To disable it please create the
/home/u/.hushlogin file.
u@Zero01:~$ systemctl status
● Zero01
    State: running
     Jobs: 0 queued
   Failed: 0 units
    Since: Tue 2023-05-23 15:19:48 -03; 11s ago
   CGroup: /
           ├─user.slice
           │ └─user-1000.slice
           │   ├─user@1000.service
           │   │ └─init.scope
           │   │   ├─797 /lib/systemd/systemd --user
           │   │   └─798 (sd-pam)
           │   └─session-c1.scope
           │     ├─623 /bin/login -f
           │     └─805 -bash
           ├─init.scope
           │ └─1 /sbin/init
           └─system.slice
             ├─systemd-networkd.service
             │ └─102 /lib/systemd/systemd-networkd
             ├─systemd-udevd.service
             │ ├─ 76 /lib/systemd/systemd-udevd
             │ ├─ 81 /lib/systemd/systemd-udevd
             │ ├─ 82 /lib/systemd/systemd-udevd
             │ ├─ 83 /lib/systemd/systemd-udevd
             │ ├─ 84 /lib/systemd/systemd-udevd
             │ ├─ 85 /lib/systemd/systemd-udevd
             │ ├─ 86 /lib/systemd/systemd-udevd
             │ ├─ 87 /lib/systemd/systemd-udevd
u@Zero01:~$ systemctl status ssh
● ssh.service - OpenBSD Secure Shell server
     Loaded: loaded (/lib/systemd/system/ssh.service; disabled; vendor preset: enabled)
     Active: inactive (dead)
       Docs: man:sshd(8)
             man:sshd_config(5)
u@Zero01:~$
```

### DIFF
--- a/DistroLauncher/ConfigRootFs.cpp
+++ b/DistroLauncher/ConfigRootFs.cpp
@@ -17,10 +17,15 @@
 
 #include "stdafx.h"
 #include "Patch.h"
-#include "ApplyConfigPatches.h"
+#include "ConfigRootfs.h"
 
 namespace Ubuntu
 {
+    void ConfigRootFs(const std::wstring& DistroName, WslApiLoader& wsl)
+    {
+        ApplyConfigPatches(DistroName);
+    }
+
     void ApplyConfigPatches(std::wstring_view DistroName)
     {
         // I think the WslPathPrefix no longer belongs to Oobe namespace. But let's leave that for later.
--- a/DistroLauncher/ConfigRootFs.cpp
+++ b/DistroLauncher/ConfigRootFs.cpp
@@ -17,7 +17,7 @@
 
 #include "stdafx.h"
 #include "Patch.h"
-#include "ConfigRootfs.h"
+#include "ConfigRootFs.h"
 
 namespace Ubuntu
 {
@@ -59,6 +59,6 @@ namespace Ubuntu
     void DisableSystemdUnits(WslApiLoader& wsl)
     {
         DWORD exitCode;
-        auto hr = wsl.WslLaunchInteractive(L"systemctl disable ssh &>/dev/null", true, &exitCode);
+        wsl.WslLaunchInteractive(L"systemctl disable ssh &>/dev/null", TRUE, &exitCode);
     }
 }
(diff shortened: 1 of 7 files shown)


EduardGomezEscandell (approved):
It is unfortunate that our original patch system cannot perform this action. Either way, this expansion is good and remains tidy 👍

PR #395, opened 2023-05-24
chore: Update UDI Submodule and Flutter SDK

CarlosNihelton:
We need the `ubuntu_logger` updates and Flutter 3.10 to allow the build workflow to succeed.

Details in:

https://github.com/canonical/ubuntu-desktop-installer/commit/62a5ead0c592cd77a2876e9634b9fa73dc9c94e6

https://github.com/canonical/ubuntu-flutter-plugins/commit/6d74e3d5699e5b31b1fd56983b71ff01ff2ac625

### DIFF
--- a/.github/workflows/build-pr.yaml
+++ b/.github/workflows/build-pr.yaml
@@ -66,7 +66,7 @@ jobs:
         if: ${{ env.withFlutterApp == 'true' }}
         with:
           channel: stable
-          flutter-version: '3.7.x'
+          flutter-version: '3.10.x'
       - name: Build Bundle
         working-directory: ${{ env.workDir }}
         run:  msbuild .\DistroLauncher.sln /t:Build /m /nr:false /p:Configuration=Release /p:AppxBundle=Always /p:AppxBundlePlatforms="x64" /p:UapAppxPackageBuildMode=SideloadOnly -verbosity:normal
--- a/.github/workflows/build-wsl.yaml
+++ b/.github/workflows/build-wsl.yaml
@@ -157,7 +157,7 @@ jobs:
         if: ${{ env.withFlutterApp == 'true' }}
         with:
           channel: stable
-          flutter-version: '3.7.x'
+          flutter-version: '3.10.x'
       - name: Build Bundle
         working-directory: ${{ env.workDir }}
         run:  msbuild .\DistroLauncher.sln /t:Build /m /nr:false /p:Configuration=Release /p:AppxBundle=Always /p:AppxBundlePlatforms="${{ env.AppxBundlePlatforms }}" /p:UapAppxPackageBuildMode=${{ env.UapAppxPackageBuildMode }} -verbosity:normal
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -54,7 +54,7 @@ jobs:
         uses: subosito/flutter-action@v2
         with:
           channel: stable
-          flutter-version: '3.7.x'
+          flutter-version: '3.10.x'
       - name: Install or update WSL
         uses: ./.github/actions/wsl-install
       - name: Build and install the appxbundle for testing


EduardGomezEscandell:
I don't quite understand flutter dependency management so excuse my questions.
Why do these breakages happen every now and then? Is there some dependency were we are not on a fixed version in our dependency chain? Or do the breakages happen even with fixed versions? Could we prevent these somehow, by fixing the version and using dependabot or similar?

EduardGomezEscandell (approved):

CarlosNihelton:
This is a very broad set of questions which I certainly cannot respond exhaustively, but let me try:

To start, let me point out that Dart dependencies are handled via _semantic versioning_. When we declare that we depend on `ubuntu_logger: ^0.0.1` we are saying that anything in t he range [0.0.1, 1.0.0) satisfies our constraints. At the same time that package will have its dependencies and finally we'll get to the Flutter SDK and the Dart language versions as the most fundamental dependencies in the whole matrix. So, for this example consider ubuntu_logger depending on `logging: ^1.0.1` and `logging_appenders: ^1.0.0.`.

Besides that, application projects usually commit the `pubspec.lock` file, which contains exact versions of all dependencies that were fetch the last time (including the transitive ones), and when the Dart package manager `pub` fetches a project dependencies it respect that file, if it sees one. It helps with reproducible builds. The problem with that file is that even a SDK patch version difference causes it to be regenerated. So, it's not that locking as it sounds. Unless we set CI to a very specific SDK version and recommend that same version to any contributor, we end up causing changes to the lock file very often in pull requests, and the end result wouldn't be much different from not having a lock file, IMHO, besides noise here and there. It might be still helpful, though, by moving the break to the pull request time.

UDI historically avoided shipping the lock file because it used to generate a lot of noise in pull requests, specially because we jumped between different versions of the SDK all the time. This is less the case nowadays, so for the WSL case it's likely worth a retry.

> Why do these breakages happen every now and then? 

R: I believe it boils down to how different people (and projects) understand what is a breaking change.

Still following our example, the `logging` package had an update to version 1.2.0 less than 24h ago (at the time I'm writing this), which introduced a new method in the `Logger` class. A new method it not necessarily a breaking change, **unless you implement** that class API. In Dart there are different mechanisms for inheritance: *mixin*, *extends* and *implements*.  The first two nicely reuse code, thus they allow implementation-inheritance, while the last (*implements*) does not reuse code, rather it's a promise that your class fulfils a certain interface, thus it is an interface-inheritance.

When the `logging_appenders` created an internal class which promises to implement the logging.`Logger` class, it allowed for automatically breaking when the new method appeared.

logging_appenders.DummyLogger
```dart
class DummyLogger implements Logger {
```  

> Is there some dependency were we are not on a fixed version in our dependency chain?

R: Considering the standard practice of working with version sets instead of fixed versions. We need to restrict ourselves to compatible versions, so we don't need to constantly update our `pubspec.yaml` files when there are dependency updates, because they are automatically picked up as long as they fit in the version constraints we set. So, with that in mind, I'd say no, there is not. But the previous answer shows that this is is iterative. As we break, we figure out more and more constraints. The `pubspec.lock` approach I mentioned above could change this a little bit.

Still following our example, @jpnurmi was the first to notice the issue, when pushing a new PR to the Ubuntu Desktop Installer repository. He even opened an issue and a PR in the `logging_appenders` package repository in GitHub to fix the issue. In the meanwhile he restricted the `logging` package version set in the ubuntu_logger package because he interpreted 1.2.0 as a breaking change (my guess, I didn't ask him).

ubuntu_logger pubspec.yaml

```diff
-  logging: ^1.0.1
+  logging: '>=1.0.1 <1.2.0'
  logging_appenders: ^1.0.0
```

> Or do the breakages happen even with fixed versions?

R: That would certainly not happen if the entire ecosystem worked with fixed versions. But that can easily lead to dependency conflicts, that the package manager does solve very easily in the current model so as long as the semantic versioning is respected. Also, there would be tons of update notifications by tools such as dependabot, which are currently interpreted as not needed. Having the lock file I mentioned would certainly prevent the break at the CI build job, and move it likely to the PR time, so as long as we set a very specific SDK (up to the patch) version. `3.10.x` is not enough.

> Could we prevent these somehow, by fixing the version and using dependabot or similar?

R: In this repository that would be more complicated, because what breaks us is a huge git submodule, instead of small pieces. UDI doesn't tag versions often. There are only two tags at the moment of this writing. We could set dependabot to constantly bring the latest main, which would be overwhelming. Inside the UDI repository, there is `renovate`, which creates pull requests every now and then with some dependency updates. But as I said, the whole ecosystem doesn't work in fixed versions, but semantic version ranges instead. As long as SV doesn't fail, the package manager `pub` resolves the exact versions. Pushing the `pubspec.lock` file may help with this by moving the break to the PR time, most likely. It might be worth a second try.

@jpnurmi feel free to amend any mistake I may have made or add more relevant information to the conversation.

jpnurmi:
Nice summary, Carlos. Yes, a lock file may protect you to some degree but comes with its own annoyances related to being regenerated when necessary. It was a disaster when Flutter for Linux was still in beta and we had to constantly hop between Flutter channels but there's a chance it will work better these days. Side-note/warning: the 3.7.x series ended at .12 and the 3.10.x series is already going at .2. :)

What comes to the break, the ability to treat any Dart 2 class as a pure interface is both a blessing and a curse. Dart 3 is addressing this by introducing class modifiers that will give control to package authors but many are upset because once authors start marking classes final, users will lose the ability to mock/fake/stub them for testing purposes, for example.

Google's logging framework used under the hood is otherwise nice but it's missing something pretty important - redirecting the output to a file. I'll take the blame for picking a 3rdparty extension that provided support for fancy rotating log files. It's not rocket science to write log files but it also didn't seem something worth investing time into when a ready-made component was available. To be fair, it's been successfully in use for two years now. :)

EduardGomezEscandell:
Thank you guys for your thorough explanations 😄. Seems like fixing the occasional breakage is the lesser evil at the time.

PR #396, opened 2023-05-29
fix: Prevents cloud-init from running once systemd is enabled

CarlosNihelton:
Patch applied to all releases, even those without the OOBE. It doesn't harm them.

Keeps a single place to remove once we find how to support cloud-init properly. Otherwise we'd need to repeat the Patch declaration for Ubuntu, Ubuntu-Preview and UbuntuDev.WslID.Dev.

See https://bugs.launchpad.net/cloud-init/+bug/2008727/comments/2 for the file path.

With that patch we get:

```
Welcome to Ubuntu Mantic Minotaur (development branch) (GNU/Linux 5.15.90.1-microsoft-standard-WSL2 x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage


This message is shown once a day. To disable it please create the
/home/u/.hushlogin file.
u@Zero01:~$ ls /etc/cloud/
cloud-init.disabled
u@Zero01:~$ systemctl is-system-running
running
u@Zero01:~$ ps aux | grep cloud-init
u            539  0.0  0.0   4024  1900 pts/0    S+   11:42   0:00 grep --color=auto cloud-init
u@Zero01:~$ ps aux | grep cloud-init
u            543  0.0  0.0   4024  1936 pts/0    S+   11:43   0:00 grep --color=auto cloud-init
u@Zero01:~$ ps aux | grep cloud-init
u            545  0.0  0.0   4024  1804 pts/0    S+   11:44   0:00 grep --color=auto cloud-init
u@Zero01:~$
logout
```

### DIFF
--- a/DistroLauncher/Patch.cpp
+++ b/DistroLauncher/Patch.cpp
@@ -106,6 +106,11 @@ namespace Ubuntu::PatchingFunctions
         output << "\naction=reboot\n";
         return true;
     }
+
+    bool Empty(std::istreambuf_iterator<char> /* unused */, std::ostream& /* unused */)
+    {
+        return true;
+    }
 }
 
 namespace Ubuntu
--- a/DistroLauncher/Patch.h
+++ b/DistroLauncher/Patch.h
@@ -127,6 +127,10 @@ namespace Ubuntu
 
         // Marks the distro as needing a reboot, but does not act on it.
         bool DeferReboot(std::istreambuf_iterator<char> input, std::ostream& output);
+
+        // Causes an empty file to be written, useful for disabling certain services such as cloud-init,
+        // where the mere presence of the empty file is enough to signal that it shouldn't run.
+        bool Empty(std::istreambuf_iterator<char> /* unused */, std::ostream& /* unused */);
     }
 
     /// Collection of the patches that must be applied to all releases.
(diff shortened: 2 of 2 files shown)


CarlosNihelton:
I wanted to add an end-to-end test case, but there is no proper way to assert that cloud-init won't run, because it's exact start time could vary and launcher commands could complete before systemd would even try to start it. Also, since it's part of subiquity snap, it'd be even more delayed, since it'll depend on the subiquity-server.service starting and requesting it. So, there is not a good story for testing this guy reliably.

didrocks (commented):
How will that work out once we rely on cloud-init (probably this cycle) to provision it, without clashing with the OOBE?

CarlosNihelton:
> How will that work out once we rely on cloud-init (probably this cycle) to provision it, without clashing with the OOBE?

By the time we get there, the OOBE will be allowed to run only run once, before systemd gets enabled, so this won't clash with it, as any other systemd unit. Yet, this will ofc need to be reverted since we'd depend on cloud-init to execute the provisioning tasks.

CarlosNihelton:
> systemd is getting enabled even before the launcher has nothing to do with it, no?

While systemd is enabled by the launcher, that will happen prior to the OOBE run, but it only takes effect after the OOBE finishes, when it will be the first (and only) time when the distro is "restarted".

CarlosNihelton:
> Can we try investigating selecting the No Cloud cloud-init backend and shipping that as part of the configuration itself so that we are more future proof with the incoming changes to enable cloud init for initial tasks provisioning?

I'll investigate that.

PR #398, opened 2023-05-30
Simplify cleanup of leftover distros in wsl-install action

EduardGomezEscandell:
For some reason these if-checks are not working adequately, so instead I opt for a ask-for-forgiveness rather than ask-for-permission approach.

You can see the leftover distros interfering here: https://github.com/ubuntu/WSL/actions/runs/5118578090/jobs/9202785080?pr=397


didrocks (approved):
Agreed on the approach

PR #399, opened 2023-05-30
Fix obscure Git error in wsl-checkout

EduardGomezEscandell:
Apparently you cannot fetch the branch you're in. This is not an issue when this action is called in a PR:
- `git clone ...` clones into main
- `git fetch ...` fetches the PR ref

However, when this action is ran on main (for instance, after merging the PR), the fetch will try to fetch main, and fail.

See here:
https://github.com/ubuntu/wsl-actions-example/actions/runs/5120947427/jobs/9208564434

See a Stackoverflow thread about this:
https://stackoverflow.com/questions/2236743/git-refusing-to-fetch-into-current-branch

### DIFF
--- a/.github/actions/wsl-checkout/action.yaml
+++ b/.github/actions/wsl-checkout/action.yaml
@@ -30,7 +30,7 @@ runs:
           git clone --quiet --depth 1 https://github.com/${{ github.repository }}.git ${{ inputs.working-dir }}
 
           cd ${{ inputs.working-dir }}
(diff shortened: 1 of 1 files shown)


CarlosNihelton (approved):
LGTM 👍🏽

CarlosNihelton:
#400 also fixes the most recent break in the build workflow.

didrocks (approved):
Interesting use case, never went through it. Thanks for the detailed explanation and reference

PR #400, opened 2023-05-30
chore: Update UDI submodule to let the window close once again

CarlosNihelton:
This update has two interesting properties:
1. We now have a lock file for the Flutter dependencies:
 
That helps with the constantly moving dependencies. In the presence of a `pubspec.lock` file, the `flutter pub get` command will fetch the exact dependencies locked in there. All the noise of getting that file updated with new dependencies are handled in the UDI repository. For now, they get updated at every pull request, besides the renovate workflow.

2. Updated yaru_window_manager dependency prevents never-closable window;

There was a bug in the dependency that allows us to control the window lifecycle programmatically, yaru_window_manager (on Windows), causing the window to never close. Not even manually. Details are documented in: https://github.com/ubuntu-flutter-community/yaru_window.dart/issues/29. With the updated submodule we get the fixed version of that dependency. One can confirm that by checking the version of the `yaru_window_manager` entry in the `ubuntu-desktop-installer\packages\ubuntu_wsl_setup\pubspec.lock`. It must show `0.1.2`.

Also, since @EduardGomezEscandell ported the systemd enablement to the patching mechanism, we get the side effect of running the OOBE before the systemd enablement takes effect, since we no longer restart the distro prior to closing the OOBE. 

With those changes in place the OOBE becomes functional again. `cloud-init` is still running after systemd enablement takes effect, but it doesn't prevent or delay the distro boot. Another task will take care of removing the snap providing the OOBE support and the reconfiguration option from the launcher CLI, so that subiquity and cloud-init services will no longer consume resources for no added benefit at this point in time.

---

[The artifacts](https://github.com/ubuntu/WSL/suites/13249167088/artifacts/722266001) of the Build WSL workflow run shown below demonstrate the end result of this pull request.

https://github.com/ubuntu/WSL/actions/runs/5124492815


EduardGomezEscandell (approved):
Two birds with one stone!

EduardGomezEscandell:
Merged it so I can start testing Ubuntu-18.04 and Ubuntu-20.04 with all the recent changes.

PR #404, opened 2023-07-10
Install debug control hook

EduardGomezEscandell:
In Microsoft’s implementation of the C++ standard library, there are some useful debug assertions, mostly regarding memory and address safety. The default behaviour of these is to promp the user with the line where the assertion failed (not very useful as it is often deep inside the standard library, but it offers some clue) and stop execution until that prompt is attended to.

This is usefull when locally debugging but annoying in automated CI, as that prompt will wait until the tests time out and the process is killed.

### DIFF
--- a/DistroLauncher/DistroLauncher.cpp
+++ b/DistroLauncher/DistroLauncher.cpp
@@ -109,6 +109,9 @@ int wmain(int argc, wchar_t const *argv[])
 
     // Initialize a vector of arguments.
     std::vector<std::wstring_view> arguments;
+    
+    arguments[5];
+
     for (int index = 1; index < argc; index += 1) {
         arguments.push_back(argv[index]);
     }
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell:
[See error message](https://github.com/ubuntu/WSL/actions/runs/5505468124/jobs/10034101400):
```
Messages:   	Unexpected error installing: [ASSERT] C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.29.30133\include\vector(1553) : Assertion failed: vector subscript out of range
```

This would have otherwise been a GUI prompt on the Azure VM, and the tests would have timed out waiting for someone to click OK.

didrocks (approved):
This sounds like this will be of a good help in the future in CI, nice that you found this!

PR #405, opened 2023-07-10
Fix update-cache

EduardGomezEscandell:
It was using an action that has not existed for a while


didrocks (approved):

didrocks:
(please note Update rootfs cache / update-cache (pull_request) Failing after 14m)

PR #412, opened 2023-11-22
ci: retry rootfs download if it fails

jibel:
Since Powershell 6+ invoke-Webrequest has a retry option. Use this option in case of failure together with the resume option.
Besides, the path to action.yaml was wrong.

UDENG-1813


jibel:
end to end tests are failing because 24.04 != 23.10

CarlosNihelton (approved):
Looks great!

End-to-end tests need some maintenance, outside of the scope of this PR.

PR #414, opened 2023-11-22
tests: fix version of the devel release

jibel:
quick'n dirty fix to the version. Ideally we use something similar to distro-info but since it's running on Windows it is not so straightforward.


didrocks (commented):
hum, I think this could be fixed when we detect a new release available via a template, the same way we do it to create the new assets, wdyt?

jibel:
indeed. The goal here is to quickly unblock e2e tests before going for a more robust solution.

didrocks (approved):
mind creating a card for it so that we don’t have to fix it again in 6 months?

jibel:
> mind creating a card for it so that we don’t have to fix it again in 6 months?

https://warthogs.atlassian.net/browse/UDENG-1816

PR #415, opened 2023-11-23
fix(e2e): Remove testCorrectReleaseRootfs test case

EduardGomezEscandell:
We deemed that manually updating this every now and then is not worth the effort.

---

UDENG-1816


jibel (approved):
Thanks!

PR #422, opened 2024-01-22
Fix help argument to not install the distro

EduardGomezEscandell:
The way command line arguments are parsed is not ideal, but that is mostly inherited from upstream.

Command `launcher.exe help` should print the help and exit, however `help` is not a real command, so it defaults to the same behaviour as `launcher.exe made-up-command`. It first installs and then fails to launch.

I fix this by checking for `help` very early on.

UDENG-276


CarlosNihelton (approved):
LGTM 👍🏽 
Thanks for the fix!

PR #426, opened 2024-01-30
feat(CI): Cache rootfs download

EduardGomezEscandell:
Until now he were downloading the rootfs every night, then using that rootfs during the day's workflows.

However, other repositories may want to download rootfs as well (looking at you, cloud init 👀), so offering an action that does that is nice. We already had this: `download-rootfs`, however it does not cache the download. This means that it can take upwards of 15 minutes per run. This explains the first commit: it adds caching to this action.

Now that donwloading is cached, we don't really need to download every night. This creates a fragile relationship between e2e testing and `update-rootfs-cache`. Instead, we can use `download-rootfs` on demand knowing that it'll only download once. This is the second commit.


.github/actions/download-rootfs/action.yaml
```diff
@@ -27,8 +27,10 @@ runs:
       uses: actions/setup-go@v3
       with:
         go-version: "1.21.4"
-    - name: Prepare environment
+    - name: Generate WSL release info
       shell: powershell
+      # Repository root
+      working-directory: ${{ github.action_path }}\..\..\..\
```
CarlosNihelton, line 33: I wonder whether wsl-builder should be placed inside `.github` or not.

CarlosNihelton (approved):
LGTM 👍🏽

PR #428, opened 2024-01-30
ci: Remove references to the OOBE and Flutter from the build steps

CarlosNihelton:
Continuing the clean-up, this PR removes all references of Flutter and the OOBE in the multiple CI steps that either builds or is aware of the existence of the OOBE or the dependency on Flutter.

In the end, all appx built from now on will look pretty much the same as shown below (no Flutter or OOBE binaries in the appx):

Captura de tela 2024-01-30 100516

Closes UDENG-1825


EduardGomezEscandell (approved):
👍

PR #429, opened 2024-01-30
feat: Remove the OOBE and distro patching from the code

CarlosNihelton:
This completes the removal of the OOBE and distro patching from the launcher.

Some end-to-end test cases will inherently be in a broken state until we have wsl-setup package with the patching feature ready and available by default.

Yet, instead of removing the end-to-end tests completely, I prefer to keep the few test cases that are not broken, leave space to allow us to create more test cases and take advantage of the fact that the preparation for test itself is also a test case; asserting that the changes introduced in a PR don't break build, that the resulting package is installable etc.

To demonstrate the issue with the end-to-end test cases, below is a broken run failing on systemd enablement assertion and a restricted run, passing because the broken tests are no longer exercised in CI.

Broken run: https://github.com/ubuntu/WSL/actions/runs/7712476125/job/21020192594?pr=429
A restricted run: https://github.com/ubuntu/WSL/actions/runs/7712932789/job/21021736003?pr=429

With the changes in place we no longer have the OOBE, nor distro patching, nor custom CLI.
Some things were still preserved:

- the recently introduced logic for the help command;
- relative binary paths for shell commands used in DistroLauncher.cpp and DistributionInfo.cpp;
- the CRT reporting hook;
- building with the latest platform version (10.0.22000.0);

This depends on #428 .

Closes UDENG-1823


EduardGomezEscandell (approved):
The OOBE really added a lot of complexity, didn't it 😅.

Sad to see it go, but happy to see the -8399 changes to code 💪.

PR #432, opened 2024-02-01
fix(actions/wsl-bash): Bash-WSL action: write temp script in the Windows filesystem

EduardGomezEscandell:
The /tmp directory is, as expected, temporary. This is broken with regular WSL (files persist), which is something we were using in this action.

However, enabling systemd makes it so the contents of /tmp  are no longer persistent.

---

This is being tested here:
https://github.com/ubuntu/wsl-actions-example/pull/6 (cloud-init workflow)


CarlosNihelton (approved):
LGTM 👍🏽

PR #436, opened 2024-02-07
Download rootfses specific to appid livebuild rootfs

didrocks:
We now have per appid livebuild produced rootfs. Adjust the paths to download them generically.

### DIFF
--- a/build-remote
+++ b/build-remote
@@ -204,7 +204,7 @@ def main():
                 shouldExit = True
             arch = os.path.basename(build.distro_arch_series_link)
             for f in build.getFileUrls():
-                if not f.endswith(".rootfs.tar.gz"):
+                if not f.endswith(".%s.rootfs.tar.gz" % selected_appID.lower()):
                     continue
                 rootfses.append("%s::%s" % (f, arch))
         if shouldExit:
--- a/build-remote
+++ b/build-remote
@@ -204,7 +204,7 @@ def main():
                 shouldExit = True
             arch = os.path.basename(build.distro_arch_series_link)
             for f in build.getFileUrls():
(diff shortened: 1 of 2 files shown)


build-remote
```diff
@@ -204,7 +204,7 @@ def main():
                 shouldExit = True
             arch = os.path.basename(build.distro_arch_series_link)
             for f in build.getFileUrls():
-                if not f.endswith(".rootfs.tar.gz"):
+                if not f.endswith(".%s.rootfs.tar.gz" % selected_appID.lower()):
```
EduardGomezEscandell, line 207:
```suggestion
                if not f.endswith(f".{selected_appID.lower()}.rootfs.tar.gz"):
```

F-strings are best-strings.

EduardGomezEscandell (approved):
Nitpick, feel free to ignore

build-remote
didrocks: Thanks! I never know about the recent python formatting! Picking your suggestion :)

jibel:
well, as a general rule, a PR shouldn't be approved if the tests are failing. Otherwise add a comment explaining why they are failing and if it is related or not to the changes.

didrocks:
For the end to end tests to pass, after investigation, we need to download the rootfs with the new naming convention. This means we need to validate from the last livecd-rootfs build in the ppa, then upload it to noble, and wait for one image from the CPC team to be built (the action is taking Ubuntu-Preview, fortunately)

didrocks:
Ok, so apart from the WSL official images, (we still need to have "Ubuntu" published there, the rest is passing. Current is broken too until we can finish the transition.

The naming here follow the new convention and we need to iterate too in the store for the on demand build. I’m merging this one to unblock the future work.

PR #441, opened 2024-02-27
CI: Always store the new build-id

didrocks:
Separate upload to the Store from storing the new version so that even manual build can be uploaded to the Store.

UDENG-2387

### DIFF
--- a/.github/workflows/build-wsl.yaml
+++ b/.github/workflows/build-wsl.yaml
@@ -116,11 +116,8 @@ jobs:
           # Download rootfses, checksum and place them at the correct place
           go build ./wsl-builder/prepare-build
           extraArgs=""
-          if [ ${{ matrix.Upload }} != "yes" ]; then
-              extraArgs="--build-id 9999"
-          fi
           if [ ${{ matrix.RootfsesChecksum }} != "yes" ]; then
-            extraArgs="${extraArgs} --no-checksum"
+            extraArgs="--no-checksum"
           fi
           archsBundle="$(./prepare-build prepare ${{ env.buildInfoPath }}/${{ matrix.AppID }}-buildid.md ${{ matrix.AppID }} ${{ matrix.Rootfses }} ${extraArgs})"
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wsl.yaml
```diff
@@ -213,35 +216,41 @@ jobs:
...
+          echo "::set-output name=has-changes::true"
+
+          if [ ${{ matrix.Upload }} != "yes" ]; then
+            echo "::notice::${{ matrix.AppID }} build ${build_id} ready for sideload or manual submission to the Microsoft Store"
+            exit 0
+          fi
 
-          if [ "${build_id}" = "0" ]; then
+          # If we are doing a first build for this distro, we don’t want to submit it to the store as first submission
+          # is manual.We let the other steps and jobs happening to save artifacts (build-id, fingerprints) of
```
CarlosNihelton, line 234: nitpick:
```suggestion
          # is manual. We let the other steps and jobs happening to save artifacts (build-id, fingerprints) of
```

.github/workflows/build-wsl.yaml
```diff
@@ -213,35 +216,41 @@ jobs:
...
+
           hasChanges="$(git diff --staged ${fingerprint_file})"
           cd -
-
           if [ -z "${hasChanges}" ]; then
             exit 0
           fi
 
-          echo "::set-output name=save-artifacts::true"
+          echo "::set-output name=has-changes::true"
```
CarlosNihelton, line 226: We should set some point to stop using `set-output` and prefer `$GITHUB_OUTPUT`, but that can be set for later.

CarlosNihelton (approved):
All good to me. Just two minor nitpicks which are not a priority.

.github/workflows/build-wsl.yaml
didrocks: Agreed but for later :)

.github/workflows/build-wsl.yaml
didrocks: Thanks!

PR #442, opened 2024-02-28
fix(ci): Remove unnecessary concurrency directive for build-wsl

EduardGomezEscandell:


didrocks (approved):
Whoever did that should be cursed forever :) (lalalalala)

PR #457, opened 2024-04-09
maint: allow download from cloud image and reenable automated build

didrocks:
* Adapt RootfsURL to current cloud-images
   Adapt it to our current cloud-images copy work, which strips the "ubuntu" prefix and replace it with "wsl" when the appid is empty.
   Support older versions too.
* Reenable automated build cron job

UDENG-2573

### DIFF
--- a/wsl-builder/common/releasesinfo.go
+++ b/wsl-builder/common/releasesinfo.go
@@ -9,7 +9,6 @@ import (
 	"encoding/csv"
 	"fmt"
 	"os"
-	"path"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


didrocks:
self hosted runner is not reachable apparently. (but this PR doesn’t affect it as long as we can build the package)

jibel (approved):

wsl-builder/common/releasesinfo.go
```diff
@@ -234,16 +233,27 @@ func (w *WslReleaseInfo) refreshedTerminalProfileID() error {
...
+		suffix = strings.TrimPrefix(strings.ToLower(w.AppID), "ubuntu")
+		if suffix == "" {
+			suffix = "wsl"
+		}
 	}
+	imageName := fmt.Sprintf("%s-%s.rootfs.tar.gz", imageBaseName, suffix)
 
-	return fmt.Sprintf("https://cloud-images.ubuntu.com/%s/current/%s-%s-wsl.%s.rootfs.tar.gz",
-		codeNameSubUri, imageBaseName, arch, strings.ToLower(w.AppID))
+	return fmt.Sprintf("https://cloud-images.ubuntu.com/wsl/%s/current/%s", w.CodeName, imageName)
```
CarlosNihelton, line 258: What about focal? It still appears at the server cloud-image location: https://cloud-images.ubuntu.com/focal/current/

image

image

jibel: The WSL releases are aligned with the official releases of Ubuntu. The latest release of version 20.04 is 20.04.6, which was released on March 23, 2023, and will be supported until April 2025. There will not be a 20.04.7 release, and we will not release a new build of WSL.

So, it is fine if it is excluded from this change.

PR #458, opened 2024-04-15
Update install-ubuntu-wsl2.md

joestephens:
The current guide references (I assume now removed) old learning points.


CarlosNihelton (approved):
Thanks for noticing this!

PR #463, opened 2024-04-22
Revert "Refresh releases and assets"

CarlosNihelton:
Reverts ubuntu/WSL#462

That was targeted to a local "trying to unblock the store" branch with another product and hacks.
Shouldn't have been merged.


CarlosNihelton:
No longer needed

PR #465, opened 2024-04-29
Update cloud-init.md

Dranduck:
Neither Ubuntu 24.04 nor Ubuntu-Preview comes with zip, tar, curl or unzip installed.

This guide is for Ubuntu. The script is not in sync with the guide.

Either merge this to guide or add note in the guide that the commands wont run from cloud-init


CarlosNihelton (approved):
Very good point! Thanks for the contribution!

PR #466, opened 2024-04-29
fix(docs): Updates cloud-init tutorial to refer to 24.04

cnihelton:
I think it's better to highlight the current LTS than promote Preview, now that we can.

Thus this replaces all Ubuntu-Preview related commands with 24.04.

https://canonical-ubuntu-wsl.readthedocs-hosted.com/en/fix-docs-cloudinit-ga/tutorials/cloud-init/

UDENG-2434

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -24,6 +24,7 @@ Jupyter
 localhost
 LTS
 npm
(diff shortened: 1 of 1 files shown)


jibel (approved):
👍 Thanks for the update

PR #472, opened 2024-05-28
Update dotnet-systemd.md

dbaduck:
code blocks needed to be on separate lines

### DIFF
--- a/docs/tutorials/dotnet-systemd.md
+++ b/docs/tutorials/dotnet-systemd.md
@@ -15,13 +15,13 @@ Systemd support is a new feature of WSL and only available on WSL version XX or 
 
 You can check your current WSL version by running:
 
-> `> wsl -- version`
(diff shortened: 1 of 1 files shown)


dbaduck:
This article is a mess.  There are graphics that show /demos/mybot/echoes and that is not what is in the code. I actually could not get the bot to be created. It should be reworked to actually work.

edibotopic:
> code blocks needed to be on separate lines

Thanks for identifying this issue and submitting the PR @dbaduck . We really appreciate your input.

> There are graphics that show /demos/mybot/echoes and that is not what is in the code. I actually could not get the bot to be created. It should be reworked to actually work.

If you encountered specific problems when following this tutorial it would be very helpful if you submitted a separate issue.
Providing us with any additional information will be a big help when we work on future versions of the tutorial.

PR #478, opened 2024-06-27
feat: Check init tasks

CarlosNihelton:
This affects only the registration time. Subsequent boots don't check for cloud-init or depend on it in any way.

If the distro launcher is used to register a WSL instance, we check if there is a `cloud-init` command in PATH, and if so we launch `cloud-init status --wait` right after registration is complete.
This will wait until cloud-init completes what's doing or report error/disabled states. By having a Win32 process mapped to that command seems enough to make WSL consider the boot is complete, thus preventing WSL boot time watchdog to kick in.

---
EDIT:
Cloud-init could have created Linux user accounts. Users could have been created in many other ways, even pre-created during image build. We expect the desired default user to be set via `/etc/wsl.conf`. That would work even without the distro launcher, so it's reasonable to assume the desired default to be present there.

We "parse" that file and, if there is the default user entry, then we set it via WSL API (which is faster than terminanting and restarting the instance). In order to leverage Windows primitives for parsing INI, we have to make a temporary copy of that file into the Windows filesystem.

If we don't find a default user set, we check if the current default user is still root, in which case we look for any non-system
user in the NSS passwd database (by launching getent), i.e. any user with the lowest UID greather than or equals to 1000 and  have a real login shell (no `/bin/sync`,`/bin/false` or `/usr/bin/nologin`).

Finally, we fallback to the interative user creation, i.e. the default upstream prompting.

---

I originally intended to put this implementation inside DistributionInfo.h/cpp to ensure less files being touched (thus less
diff compared to upstream) but that doesn't play nice with CI 'Refresh releases and assets' workflow.

So I placed the code in its own component, and inside a subdirectory, so if there comes a time we need to revert that change (it becomes obsolete or so) it remains easy to drop. Also, being inside a subdirectory allows me to drop a .clang-format in
there without worrying of reformatting upstream code :)

Lessons learned from the launcher's old times:
- We can put cpp files inside a subdirectory and still work with precompiled headers in Visual Studio if we either adjust the include path for every cpp file or adjust the path to the precompiled header for every cpp file in the vcxproj file (no option is winner :) )

I didn't add too many test cases because I was afraid of taking too long to register/unregister in between tests.
I also avoided adding stuff that could make the tests brittle such as installing huge apt packages (in fact I'm installing nothing at all at this moment), attaching to Ubuntu Pro, enrolling into Landscape and what not. Up for discussion on what makes sense to add here. I also didn't play with cloud-init user-data file selection. Although I could, I believe that's better tested as a unit test of the datasource in the upstream repository itself. Up for discussion as well.

Finally we cannot use private directories to behave as `%USERPROFILE%`, we need to put stuff inside the real user profile directory on Windows, which is sad. We can set `%USERPROFILE%` in the Windows shell appropriately and subprocesses of the launcher (either Linux or Windows processes) find the correct value, but systemd services won't. They will still match up the global `%USERPROFILE%`, one reason why I struggled to implement those tests.

---
UDENG-2768


didrocks (changes requested):
Before going on with the PR, I think we should clarify this:

> If we don't find a default user set, than we allow fallback to the interative user creation, i.e. the default upstream prompting.

I think we said that if we don’t find anything in wsl.conf, we still list all users and take the first one with UID >= 1000? I can imagine people using cloud-init use the standard user creation creation but don’t know about wsl.conf and we want to support them.

My suggestion would be, once the launcher waited on cloud-init as you did in that PR:
- look at `wsl.conf`, if there is a value there, ensure the user exists on the system (with `getent passwd`), update the windows registry
- if nothing, `getent passwd` list all users with UID >= 1000. If found, update wsl.conf and the windows registry
- otherwise, fallback to interactive user creation.

Does this make sense?

CarlosNihelton:
> Before going on with the PR, I think we should clarify this:
> 
> > If we don't find a default user set, than we allow fallback to the interative user creation, i.e. the default upstream prompting.
> 
> I think we said that if we don’t find anything in wsl.conf, we still list all users and take the first one with UID >= 1000? I can imagine people using cloud-init use the standard user creation creation but don’t know about wsl.conf and we want to support them.
> 
> My suggestion would be, once the launcher waited on cloud-init as you did in that PR:
> 
> * look at `wsl.conf`, if there is a value there, ensure the user exists on the system (with `getent passwd`), update the windows registry
> * if nothing, `getent passwd` list all users with UID >= 1000. If found, update wsl.conf and the windows registry
> * otherwise, fallback to interactive user creation.
> 
> Does this make sense?

While it does make sense, I've been resisting to move towards that direction because it becomes harder (IMHO) to explain to users why certain cloud-init user-data behaves in one way when the distro is set up via the launcher and differently when through `wsl --import` for example. To date I still struggle to find a clear way to present this difference in behavior as a feature of the distro launcher so users won't just assume coming straight from cloud-init and being surprised when it doesn't happen, other than relying on Ubuntu WSL documentation. I think that can be confusing. Do you still believe it's worth investing in doing this augmentation of cloud-init in the distro launcher?

didrocks:
It does not depend on cloud-init directly as we don’t really care the way the user was created on the machine. I can be a pre-created image with a default user and they didn’t create/know about `wsl.conf`.

We don’t want those users to grow that knowledge and the launcher should do the right thing silently.

@jibel opinions?

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,190 @@
...
+        // found another section, so value not found.
+        return {};
+      }
+      if (auto [k, v] = matchKeyValuePair(line); k == key) {
+        return std::string{v};
+      }
+    }
+
+    return {};
+  }
```
CarlosNihelton, line 106: This stops on the first occurence of `[section].key`. For a moment I considered that could be a mistake, but indeed that's how WSL parses ini:

image

I'll add one comment about this for the future.

jibel:
> It does not depend on cloud-init directly as we don’t really care the way the user was created on the machine. I can be a pre-created image with a default user and they didn’t create/know about `wsl.conf`.
> 
> We don’t want those users to grow that knowledge and the launcher should do the right thing silently.
> 
> @jibel opinions?

Looking at the order of priority to consider a default user for WSL it's:
wsl.conf > registry 
wsl considers nothing else.

So we should check:
* wsl.conf
* registry
* first user >= 1000 (possibly created by cloud-init or a wsl import or something else we don't know)
* interactive mode

Then create an entry in wsl.conf based on this discovery if there is none so wsl -d starts with the default user

CarlosNihelton:
I changed the PR description to match the current implementation after our discussion. The changes in the description are marked  as shown below:

---

EDIT:
...

---

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+    if (found != users.end()) {
+      writeUserToWslConf(found->name);
+      return true;
+    }
+  }
+
+  // No default user set anywhere. Let's search for the first non-system user.
+  // UID 65534 is typically the user 'nobody' and non-system users have UID>=1000.
+  auto found = std::find_if(users.begin(), users.end(),
+                            [](const UserEntry& u) { return u.uid > 999 && u.uid < 65534; });
```
didrocks, line 155: I don’t think going with u.uid < 65534 is a good one, We can skip the nobody user, but all Active Directory users, or Azure Active Directory will have a uid > 65534.

Another way to filter invalid users would be to take the shell argument on them, and exclude those having `/usr/sbin/nologin` or `/bin/false` (which should cover 99% of the spectrum).

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+// Collects all users found in the NSS passwd database, sorted by UID.
+std::vector<UserEntry> getAllUsers(WslApiLoader& api);
+
+// Returns the defaultUser set in /etc/wsl.conf or the empty string if none is set.
+std::string defaultUserInWslConf();
+
+// Converts a multi-byte into a wide string.
+std::wstring str2wide(const std::string& str, UINT codePage = CP_THREAD_ACP);
+
+bool enforceDefaultUser(WslApiLoader& api) try {
```
didrocks, line 122: it’s probably me, but I didn’t see the case which comes in second which is "there is a key registry, then we consider this was the default set for that distro?" as the order is wsl.conf > registry?

e2e/launchertester/basic_setup_test.go
```diff
@@ -18,20 +20,73 @@ func TestBasicSetup(t *testing.T) {
...
+// TestSetupWithCloudInit runs a battery of assertions after installing with the distro launcher and cloud-init.
+func TestSetupWithCloudInit(t *testing.T) {
+	testCases := map[string]struct {
+		install_root bool
+		wantUser     string
+		wantFile     string
+	}{
+		"With default user in conf":    {wantUser: "testuser", wantFile: "/etc/with_default_user.done"},
+		"Without default user in conf": {wantUser: "testuser", wantFile: "/home/testuser/with_default_user.done"},
+		"Without checking user":        {install_root: true, wantUser: "root", wantFile: "/home/testuser/with_default_user.done"},
```
didrocks, line 47: It’s not in this test per say, but there is probably one when there is a default user set in the registry, with or without wsl.conf.

didrocks (changes requested):
I think we are getting there! I still see some discrepancy with what was discussed on MM though which I think we should fix before merging:
1. We told that the order is (from what wsl command is considering):
- wsl.conf
- key in registry

If I am correct, we don’t check the registry here, only set it. I think similarly than wsl.conf, we should consider it as "a default has been set".

2. You are considering a range rather than user >= 1000. I think you are right to not consider the nobody user, but there are still some other potentials users with a way larger uid when they are registered through Active Directory for instance or any other dynamic "bindings". I think the approach to ignore some users is correct and we should filter by the shell associated with it.

3. I see that you try to "fix" any invalid wsl.conf (and if the registry is checked, I think you would do the same). I am more in favor of "so be it". If wsl.conf or the registry are configured, that was an intent by the rootfs owner or by the cloud-init which was set. If this is set, let’s not do anything, no double checking, no fix and error out if this is incorrect.
That way, we don’t diverge from "we have a cloud-init" behavior to "there is none" with the wsl command itself?

So in a nutshell, I think we should:
- if there is `wsl.conf` user entry -> no additional check, no user creation, let’s then wsl run with this user (even if it doesn’t exist)
- else if there is a registry key -> no additional check, no user creation, let’s then wsl run with this user (even if it doesn’t exist)
- else if there is one: get first user with uid >= 1000 and an associated shell (no /bin/false, no /usr/sbin/nologin) -> set it as default in wsl.conf (do we need to update the registry if wsl.conf always takes precedence?)
- else -> default to interactive prompting

Does it make sense?


NB: I feel sorry you had to write our own ini parser here. I tried to review the C++ part to the best of my knowledge but I fully trust you on that side :)

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+  if (auto name = defaultUserInWslConf(); !name.empty()) {
+    auto found = std::find_if(users.begin(), users.end(),
+                              [&name](const UserEntry& u) { return u.name == name; });
+    if (found != users.end()) {
+      return setDefaultUserViaWslApi(api, found->uid);
+    }
+    wprintf(
+        L"CheckInitTasks: user \"%s\" referred in /etc/wsl.conf "
+        L"was not found in the system, entry may be overwritten.\n",
+        str2wide(name).c_str());
```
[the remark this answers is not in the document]
jibel: Agreed. The image (not an official ubuntu image) can have been customised this way for a reason we ignore.
If the user defined in wsl.conf doesn't exist, it uses the user defined in the registry and if the user in the registry doesn't exist either, it falls back too root with the following error:

```
<3>WSL (570) ERROR: CreateProcessParseCommon:748: getpwnam(xxx) failed 17
<3>WSL (570) ERROR: CreateProcessParseCommon:757: getpwuid(1002) failed 17
```

But the session starts anyway.

DistroLauncher/Ubuntu/InitTasks.cpp
jibel: Actually if you want to check a range on Ubuntu the default for UIDs and GIDs is [1000, 60000]

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+  return false;
+} catch (...) {
+  _putws(L"CheckInitTasks: Unexpected failure when trying to set the default WSL user");
+  return false;
+}
+
+// Decomposes a line such as `key = value` into its [key] and [value] components.
+std::pair<std::string_view, std::string_view> matchKeyValuePair(std::string_view line);
+
+// A limited INI "parser" capable of retrieving the value of a [section].key entry.
```
jibel, line 173: Not a Windows dev expert here, but why do not you use the Windows primitive to read/write the ini file?
For reference: https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getprivateprofilestring

It would reduce the size of the code.

CarlosNihelton:
Let me interject on this:

> We told that the order is (from what wsl command is considering):
wsl.conf
key in registry
If I am correct, we don’t check the registry here, only set it. I think similarly than wsl.conf, we should consider it as "a default has been set".

The function `bool enforceDefaultUser(WslApiLoader& api)` does the following order:
1. stores all user entries from `getent` (used for name/uid matching in the subsequent steps)
2. reads /etc/wsl.conf
3. checks if there is one default set via WSL API (which is the same as checking the registry -- via `if (auto uid = DistributionInfo::QueryUid(L""); uid != 0) {`
4. finally searches for a non-system user in the collection stored in 1.

Step 3 needs a comment in the code to clarify that behavior, but in fact it's checking for the registry by its effects.
Under the hood `DistributionInfo::QueryUid(L"")` runs `id -u `  (without specifying the username), effectively checking the uid of the current user running that command, i.e. the current default user.

If one manually sets the relevant registry key or runs `WslConfigureDistribution(uid, flags)`, the effect is immediate, i.e. the next WslLaunch or WslLaunchInteractive will run commands as that user.
`/etc/wsl.conf` is more problematic. Changes in `/etc/wsl.conf` only become effective **upon instance restart** (not the VM, but the particular instance).

So, consider that a rootfs comes with `boot.systemd=true` out of the shelf. The first command we'd launch after registering an instance from that rootfs will have systemd running, because during boot WSL already sees that definition. Default user by then is root (nothing set the default user yet). Let's assume the first command is then: `cloud-init status --wait` and, by the time cloud-init finishes, it writes `user.default=ubuntu` into `/etc/wsl.conf`. The default user will remain root, because the instance was not rebooted yet. That's why when we read `/etc/wsl.conf` right after cloud-init finishes its job we set the default user via WSL API (or reboot the instance - not the VM, just the instance).

I'll clarify that in the code, but otherwise I don't believe we need to read the registry directly. Even because that would be more prone to races than what we're doing now.

CarlosNihelton:
About us writing into `/etc/wsl.conf` it was more for consistency, as proposed by jibel on this comment:

> So we should check:
> 
> wsl.conf
> registry
> first user >= 1000 (possibly created by cloud-init or a wsl import or something else we don't know)
> interactive mode
> Then create an entry in wsl.conf based on this discovery if there is none so wsl -d starts with the default user


I think we can even skip writing that file, because if we set the default user via WSL API (which is the same as writing into the registry) then `wsl -d` runs with the default user set unless specified otherwise by the `u user` option.

By avoiding writing it we avoid "fixing" it at the same time. What do you guys think? @didrocks @jibel

didrocks:
Ack on reading the API indirectly from the registry (yes, a comment on this will help) :)

> I think we can even skip writing that file, because if we set the default user via WSL API (which is the same as writing into the registry) then wsl -d runs with the default user set unless specified otherwise by the u user option.

If I am correct, this is still checking the user exists and fix the consistency issue no? Again, I’m not close to the code, but I think basically the source of truth should be:

1. if wsl.conf is there and wrong (by wrong, I mean the user doesn’t exist), so be it, let’s the already set value in it
2. if the registry is there and wrong, same paradigm, we don’t do any consistency check and don’t enforce/overwrite ourself.

Basically, we only set one of the two, depending on what you told before, if both of them are not set. This is my 2 cents.

CarlosNihelton:
> Ack on reading the API indirectly from the registry (yes, a comment on this will help to clarify it) :)
> 
> > I think we can even skip writing that file, because if we set the default user via WSL API (which is the same as writing into the registry) then wsl -d runs with the default user set unless specified otherwise by the u user option.
> 
> If I am correct, this is still checking the user exists and fix the consistency issue no by changing the registry? Again, I’m not close to the code, but I think basically the source of truth should be:
> 
> 1. if wsl.conf is there and wrong (by wrong, I mean the user doesn’t exist), so be it, let’s the already set value in it
> 2. if the registry is there and wrong, same paradigm, we don’t do any consistency check and don’t enforce/overwrite ourself.
> 
> Basically, we only update one or the two of them, depending on what you told before, if both of them are unset. This is my 2 cents.

I totally agree on error out if `/etc/wsl.conf` is wrong.
We still have an issue if that file is correct: we'll start the shell as root, because the instance didn't reboot. Setting the UID matching the name in `/etc/wsl.conf` would prevent that behavior. Otherwise we must restart the instance, which is slower, but allows any change in that conf file to take effect (which might be more than just setting the default user in the end).

jibel:
> About us writing into `/etc/wsl.conf` it was more for consistency, as proposed by jibel on this comment:
> 
> > So we should check:
> > wsl.conf
> > registry
> > first user >= 1000 (possibly created by cloud-init or a wsl import or something else we don't know)
> > interactive mode
> > Then create an entry in wsl.conf based on this discovery if there is none so wsl -d starts with the default user
> 
> I think we can even skip writing that file, because if we set the default user via WSL API (which is the same as writing into the registry) then `wsl -d` runs with the default user set unless specified otherwise by the `u user` option.
> 
> By avoiding writing it we avoid "fixing" it at the same time. What do you guys think? @didrocks @jibel

I think it is te correct behaviour. If one user is created by cloud-init and at the same time wsl.conf is created from cloud-init but with a different user, we would have the first user in the registry, but wsl.conf would take the precedence and use another user. This is the behaviour we want.

What happends if several users are created by cloud-init? Which one would be in the registry?

CarlosNihelton:
> What happends if several users are created by cloud-init? Which one would be in the registry?

Assuming nothing else sets the default user by then, the launcher will iterate over the list of users collected from the passwd database and set as default the **lowest UID greater than 1000**, which matches the first user written in the cloud-config user-data, per my observation when writing tests (assuming only cloud-init would create non-system users in that scenario).

DistroLauncher/Ubuntu/InitTasks.cpp
CarlosNihelton: That INI-related Win32 functions works will all sorts of local paths, with or without Windows line terminations, but unfortunately not with WSL paths (or network paths in general I assume).

You can check a proof-of-concept [on GitHub](https://github.com/CarlosNihelton/pocs/tree/main/IniS2)

If we really want to leverage it. we could copy the file to some temporary location on Windows to parse it using that function. That will certainly be shorter, though it seems a bit strange to me having to copy a file to another location to be able to read it.

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+struct UserEntry {
+  std::string name;
+  ULONG uid = -1;
+};
+
+namespace fs = std::filesystem;
+fs::path wslConfPath() {
+  // init-once, lazily.
+  static fs::path etcWslConf =
+      fs::path{L"\\\\wsl.localhost"} / DistributionInfo::Name / "etc\\wsl.conf";
```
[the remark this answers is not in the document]
CarlosNihelton: As long as the first argument is a `std::filesystem::path` object the `operator/` performs a concatenation of path segments using the native path separator, being the segments  `std::filesystem::path` or string-like things that are convertible to  `std::filesystem::path`. One extra care we need to take when doing this kind of concatenation is that the right hand side elements must not contain a root segment, otherwise the concatenation becomes replacement.

`"C:\\foo" / "/bar"` => `C:\\foo\\bar` but `"C:\\foo" / "D:\\bar"` => `D:\\bar` on Windows.

That behavior doesn't affect us because `DistributionInfo::Name` and `"etc\\wsl.conf"` won't contain root elements.

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+namespace {
+// Blocks the current thread until all initialization tasks finish.
+void waitForInitTasks(WslApiLoader& api);
+
+// Enforces the existence of a default WSL user either:
+// - defined in /etc/wsl.conf (which might not be in effect yet)
+// - defined in WSL API/registry
+// - or the lowest UID >= 1000 in the NSS passwd database
+// Returns false if a default user couldn't be set.
+bool enforceDefaultUser(WslApiLoader& api);
```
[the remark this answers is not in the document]
CarlosNihelton: The enforcement of declaration order is a feature of C-derived languages :)

You'll see me moving declarations around exactly for that. We could just declare all the things on top (is one way), but I like to keep declarations closer to their usage as much as possible, otherwise they just become noise.

DistroLauncher/Ubuntu/InitTasks.cpp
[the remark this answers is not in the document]
CarlosNihelton: Agreed

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+        L"CheckInitTasks: user \"%s\" referred in /etc/wsl.conf "
+        L"was not found in the system, entry may be overwritten.\n",
+        str2wide(name).c_str());
+  }
+
+  if (auto uid = DistributionInfo::QueryUid(L""); uid != 0) {
+    auto found = std::find_if(users.begin(), users.end(),
+                              [&uid](const UserEntry& u) { return u.uid == uid; });
+    if (found != users.end()) {
+      writeUserToWslConf(found->name);
```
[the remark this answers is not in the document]
CarlosNihelton: We'd not need to do so because it would just be setting what is already set, i.e. if `DistributionInfo::QueryUid(L"")` returns something other than 0, it's because somehow the default user was set via WSL API or directly in the registry.

DistroLauncher/Ubuntu/InitTasks.cpp
CarlosNihelton: Considering that eventually LDAP and other means may become enabled in WSL, I believe checking the login shell is the best approach indeed.

DistroLauncher/Ubuntu/InitTasks.cpp
CarlosNihelton: Added a comment to clarify that. We check indirectly by calling ´DistributionInfo::QueryUid(L"")` which evaluates the effect, i.e. it'd return the UID set via registry or WSL API.

e2e/launchertester/basic_setup_test.go
CarlosNihelton: That was tricky. I'm still afraid of races, but let's see what CI thinks of my approach to test that.

e2e/launchertester/basic_setup_test.go
```diff
@@ -80,9 +86,70 @@ func TestSetupWithCloudInit(t *testing.T) {
...
+							time.Sleep(300 * time.Millisecond)
+							continue
+						}
+						uid, err := strconv.Atoi(strings.TrimSpace(string(out)))
+						if err != nil {
+							t.Logf("Failed to parse %s: %v", out, err)
+							time.Sleep(300 * time.Millisecond)
+							continue
+						}
+						if uid < 1000 || uid > 60000 {
```
didrocks, line 138: hum, the condition seems to be linked to the previous case, no?

didrocks (changes requested):
Nice work! I really prefer this implementation which is easier to track (partially because of the ini parsing removal, but the comments helps too!)

I only have some leftover comments on the tests themselves:
* I think some condition are coming from the previous implementation
* Also, I feel that maybe we can augment the tests with pre-shipped /etc/passwd, where we have users with uid, but no shell, and various corner cases like this.

Do you mind investigating those?

Note: I don’t think the edit of the description is up to date though. You are still saying:
"user in the NSS passwd database (by launching getent), i.e. any user with UID between 1000 and 65534 (the nobody)."

Also, please remember that the description is used as the commit content of the project, so some care is needed before doing the finale merge to remove the old description to not mislead the user. :)

e2e/launchertester/basic_setup_test.go
CarlosNihelton: Not exactly. For this test case we rely on cloud-init to create users and the cloud-config used is under control of the test case, so it seems reasonable to assume the UID we want to set from outside will be in that range.

didrocks: I would then set that this is a setup issue? To clarify it’s not something else that failed in the test itself.

DistroLauncher/Ubuntu/InitTasks.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+  return enforceDefaultUser(api);
+}
+
+namespace {
+void waitForInitTasks(WslApiLoader& api) {
+  _putws(L"Checking for initialization tasks...\n");
+
+  DWORD exitCode_ = -1;
+  if (auto hr = api.WslLaunchInteractive(L"which -s cloud-init", FALSE, &exitCode_); FAILED(hr)) {
```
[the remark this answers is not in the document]
CarlosNihelton: Ok. We no longer check, but I made a little customization to avoid printing `/bin/bash error: command cloud-init not found` (or similar :) )

CarlosNihelton:
> I think some condition are coming from the previous implementation
Replied inline, the condition has meaning to the current test cases leveraging it.

> Also, I feel that maybe we can augment the tests with pre-shipped /etc/passwd, where we have users with uid, but no shell, and various corner cases like this.

I added a few more test cases in which cloud-init plays a double-agent role, doing its expected job but also deliverying broken `/etc/passwd` files that could mess with our logic.

> Note: I don’t think the edit of the description is up to date though. You are still saying:
"user in the NSS passwd database (by launching getent), i.e. any user with UID between 1000 and 65534 (the nobody)."

I updated the PR description.

CarlosNihelton:
Looks like I need to rebase due doc changes. I'm marking the new commits as usual.

CarlosNihelton:
> Looks like I need to rebase due doc changes. I'm marking the new commits as usual.

Rebasing was not exactly needed nor was that the fix, but rather telling bash.exe about handling symlinks on WIndows.
image

didrocks (approved):
Excellent! Thanks for working on those tests! :)

PR #479, opened 2024-07-05
docs: Add contributing docs

edibotopic:
This adds contribution guidelines in the following way:

- `CONTRIBUTIONS.md` added to root of repo
- Content from `CONTRIBUTIONS.md` included in `docs/guides/contributing.md`
- Reference to `docs/guides/contributing.md` added to homepage
- Links relevant to contributing added to homepage
- Additional minor changes (e.g., modified custom wordlist)

Note, the contributing guidelines refer to both Ubuntu WSL and Ubuntu Pro for WSL, due to the expected merging of these doc sets.

Some technical input might be needed for material relating to how contributors can contribute to the Ubuntu WSL code base specifically.

UDENG-3268

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,202 @@
+# Contributing
+
+<!-- Include start contributing intro -->
+
+## Ubuntu WSL and Ubuntu Pro for WSL
+
+The documentation for Ubuntu WSL and Ubuntu Pro for WSL are currently
+in the process of being merged. These contribution guidelines refer
+to both products.
+
+Contributions to Ubuntu WSL and Ubuntu Pro for WSL are welcomed and encouraged.
+
+## Guidelines for your contributions
+
+To ensure that making a contribution is a positive experience for both contributor and reviewer we ask that you read and follow these community guidelines.
+
+This communicates that you respect our time as developers. We will return
+that respect by addressing your issues, assessing proposed changes and
+finalising your pull requests, as helpfully and efficiently as possible.
+
+These are mostly guidelines, not rules. Use your best judgement and feel free to
+propose changes to this document in a pull request.
+
+<!-- Include end contributing intro -->
+
+## Quicklinks
+
+* [Code of conduct](#code-of-conduct)
+* [Getting started](#getting-started)
+* [Issues](#issues)
+* [Pull requests](#pull-requests)
+* [Contributing to the code](#contributing-to-the-code)
+* [Contributing to the docs](#contributing-to-the-docs)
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,202 @@
...
+### How to contribute
+
+Contributions are made via Issues and Pull Requests (PRs). A few general guidelines that cover both:
+
+* Use the advisories page of the repository and not a public bug report to
+report security vulnerabilities. For Ubuntu WSL, please use [launchpad private
+bugs](https://bugs.launchpad.net/ubuntuwsl/+filebug) which is
+monitored by our security team. On an Ubuntu machine, it’s best to use
+`ubuntu-bug wsl` to collect relevant information.
+<!-- QUERY: is `wsl` the correct argument here???. -->
```
jibel, line 75: Actually we cannot recommend this. It would:
1. require that the user is using an Ubuntu machine as you mention, which we cannot expect since they are likely running Windows.
2. report a bug to wsl in Launchpad which is Wsman Shell command Line. So totally unrelated to our WSL (WSL is also the World Surf League but nothing to do with our topic ;) )

I would recommend to report issues directly to the GitHub project.

jibel (changes requested):
minor change on reporting bugs

CONTRIBUTING.md
edibotopic: That's fine, I can recommend reporting issues directly on GitHub.

I thought the original wording might need to be removed. Initially, I had `ubuntu-bug ubuntuwsl`, following the launchpad link above but that did not seem work. I wasn't aware of Wsman Shell.

Also, I thought in this case "ubuntu machine" would include ubuntu running on WSL, because the `ubuntu-bug` command can be run in Ubuntu WSL.

CarlosNihelton:
CI failure is unrelated to this work, but more fundamentally end-to-end tests don't have to run when we only change documentation.
It turns out the workflow is defined to skip if the changes only happen inside the `/docs` directory.
In this case it was triggered because `CONTRIBUTING.md`  is in the repository root.
I'd suggest to add another entry to `paths-ignore` to skip that workflow on any markdown file in the repository root, such as:

```diff
name: End-to-End Tests
on:
  workflow_dispatch:
  pull_request:
    paths-ignore:
      - docs/**
+    - *.md  
```

edibotopic:
That makes sense @CarlosNihelton and I would support that change. Want me to do that in this PR?

CarlosNihelton (approved):
LGTM 👍🏽

jibel (approved):
Great update, thanks!

PR #491, opened 2024-08-22
fix: More silent cloud-init runs

CarlosNihelton:
The launcher focus is just protecting cloud-init run. Let's not print a word about it.
At the surface it doesn't seem that anything is different from the original UX, except for the potentially longer setup time.

We rely on wsl-setup and motd to deliver a one-time show warning message if cloud-init reports error (recoverable or not).
See: https://github.com/ubuntu/wsl-setup/pull/14 for more context about this part.

Apart from messaging, no other behaviour is changed.

Considering a condition of partial failure for cloud-init (and experience has shown there are tons of ways to go into a degraded done extended status), we move from this

image

Into this:

image

---
Closes UDENG-4111


didrocks (approved):
Will give a better user experience :)

PR #492, opened 2024-08-29
docs: update installation methods

edibotopic:
This removes the winget method from the installation instructions because it is not currently well-supported.

It is replaced by a section on installing without the Microsoft Store, as this is important to some users.
However, it is made clear that we do not currently recommend a specific store from which to download images.

A brief section on running an instance of Ubuntu was also updated, as it was previously tied to the winget section and did not use the latest LTS.

UDENG-4097

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -1,3 +1,4 @@
+Appx
 AppxPackage
 awk
 backports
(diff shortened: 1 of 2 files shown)


CarlosNihelton (approved):
LGTM 🚀

jibel (approved):

PR #496, opened 2024-09-18
e2e: add smoke test for running snaps

zyga:
As a part of cross-distro effort, we are trying to make sure that snapd works well in WSL environments. We now have WSL-specific release tests that check new snapd releases in Ubuntu 22.04 and Ubuntu 24.04 as running inside WSL. We want to complement this with a small smoke test that ensures subsequent WSL releases do not regress running snapd.

The test itself installs the hello snap from the store and runs it.

Jira: https://warthogs.atlassian.net/browse/SNAPDENG-24157


zyga:
I've closed this in favor of https://github.com/ubuntu/WSL/pull/497 which is opened from the same repository.

PR #497, opened 2024-09-18
e2e: add smoke test for running snaps

zyga:
As a part of cross-distro effort, we are trying to make sure that snapd works well in WSL environments. We now have WSL-specific release tests that check new snapd releases in Ubuntu 22.04 and Ubuntu 24.04 as running inside WSL. We want to complement this with a small smoke test that ensures subsequent WSL releases do not regress running snapd.

The test itself installs the hello snap from the store and runs it.

Jira: https://warthogs.atlassian.net/browse/SNAPDENG-24157

### DIFF
--- a/e2e/launchertester/basic_setup_test.go
+++ b/e2e/launchertester/basic_setup_test.go
@@ -32,6 +32,7 @@ func TestBasicSetup(t *testing.T) {
 		"UpgradePolicyIdempotent": testUpgradePolicyIdempotent,
 		"InteropIsEnabled":        testInteropIsEnabled,
 		"HelpFlag":                testHelpFlag,
+		"SnapdWorks":              testSnapdWorks,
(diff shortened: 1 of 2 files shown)


CarlosNihelton (approved):
Thanks for the contribution!

I reckon relying on network to download the snaps might become flaky, but for now let's keep it simple and, if needed, we investigate later how to not depend on it.

CarlosNihelton:
For the record, some unrelated part of the CI is failing due the deprecation of the `actions/upload-artifact@v2` (the same is also true for the download counterpart).

zyga:
We could explore a follow-up that uses github actions to cache downloaded snaps (snapd, core22 and hello) as well as their assertions, inject them into the WSL side and make the test network-free in the case of hot cache.

PR #500, opened 2024-10-01
docs: Update cloud-init tutorial for 24.04.1

CarlosNihelton:
With that release comes changes to the distro launcher making it no longer necessary for users to run cloud-init commands. The distro launcher does it, so the tutorial can be made shorter and more direct.

Also there was a caveat about `wsl --install` that's no longer applicable, as that installation command also goes through the launcher, both workflows are now consistent.

PR #501, opened 2024-10-03
docs: Remove autoinstall doc

edibotopic:
This removes documentation of the autoinstall feature (deprecated) and redirects the old url for autoinstall to the cloud-init doc. Follow up to discussion in #500 .


CarlosNihelton (approved):
🚀

PR #508, opened 2024-12-16
Fix cloud-init blocking on WSL1

CarlosNihelton:
`cloud-init status --wait` blocks forever if cloud-init won't ever start. Known cases where that may happen:
1. the cloud-init systemd units being masked or disabled by default in the rootfs (not our case)
1. systemd not enabled by default in the rootfs image (not our case either)
1. the distro instance being registered as a WSL 1 instance (we don't control that :) )

We're not considering that properly. With the patch https://github.com/ubuntu/WSL/pull/508/commits/bb7265201cb521ec23a636e4d7f221376880eb4c, we check if both the cloud-init.service unit is enabled (which covers 1) as well as systemd not offline, which covers 2. and 3.
The very same check was implemented in the new script that will replace this launch in the new WSL format via this PR: https://github.com/ubuntu/wsl-setup/pull/17.

I also added an end-to-end test case that sets WSL1 as the default setting globally and attempts to register the instance. Without that patch the test would timeout. See https://github.com/ubuntu/WSL/pull/508/commits/472a2d8d65db263385bd69c79f2bdfeef0ed8b5c.

UDENG-5629.

---

CI had two issues blocking me:

1. `releases-info`, a Go binary used to select which rootfses to build the application packages from, had outdated considerations about what CPC publishes under https://cloud-images.ubuntu.com/wsl:
- Jammy file naming conventions are now the same as Noble's.
- The CPC publisher no longer strips the "ubuntu" suffix on both releases.

Look at https://cloud-images.ubuntu.com/wsl/jammy/current/:

```
...
[   ] SHA256SUMS.gpg                                      2024-12-13 15:01  833
[   ] ubuntu-jammy-wsl-amd64-ubuntu.rootfs.tar.gz         2024-12-12 21:42  324M  File system image and Kernel packed
[   ] ubuntu-jammy-wsl-amd64-ubuntu22.04lts.rootfs.tar.gz 2024-12-12 21:42  324M  File system image and Kernel packed
[TXT] ubuntu-jammy-wsl-amd64-wsl.manifest                 2024-12-12 21:42   17K  Package manifest file
[   ] ubuntu-jammy-wsl-arm64-ubuntu.rootfs.tar.gz         2024-12-12 21:45  308M  File system image and Kernel packed
[   ] ubuntu-jammy-wsl-arm64-ubuntu22.04lts.rootfs.tar.gz 2024-12-12 21:45  308M  File system image and Kernel packed
[TXT] ubuntu-jammy-wsl-arm64-wsl.manifest                 2024-12-12 21:45   17K  Package manifest file
...
```

The images contain a suffix "ubuntu22.04lts".
Also, that naming convention used to be from 24.04 onwards.
But as we can clearly see 22.04 and 24.04 are following the same naming convention
(see https://cloud-images.ubuntu.com/wsl/noble/current/):

```
[   ] SHA256SUMS.gpg                                      2024-12-13 05:37  833
[   ] ubuntu-noble-wsl-amd64-ubuntu.rootfs.tar.gz         2024-12-12 20:21  347M  File system image and Kernel packed
[   ] ubuntu-noble-wsl-amd64-ubuntu24.04lts.rootfs.tar.gz 2024-12-12 20:21  347M  File system image and Kernel packed
[TXT] ubuntu-noble-wsl-amd64-wsl.manifest                 2024-12-12 20:20   16K  Package manifest file
[   ] ubuntu-noble-wsl-arm64-ubuntu.rootfs.tar.gz         2024-12-12 20:23  332M  File system image and Kernel packed
[   ] ubuntu-noble-wsl-arm64-ubuntu24.04lts.rootfs.tar.gz 2024-12-12 20:23  332M  File system image and Kernel packed
[TXT] ubuntu-noble-wsl-arm64-wsl.manifest                 2024-12-12 20:23   16K  Package manifest file

```

2. Every now and then the PowerShell script that downloads the rootfs and validate the checksums failed with a shallow stack trace of an attempt to index into a null array.
That was not really helpful. After investing some time on this I could finally learn that we implemented the wrong regex pattern (two spaces instead of a space and a raw star/asterisk character).

### DIFF
--- a/.github/actions/download-rootfs/download-rootfs.ps1
+++ b/.github/actions/download-rootfs/download-rootfs.ps1
@@ -70,7 +70,7 @@ function Test-Checksums {
         return $false
     }
 
-    $newSHA256 = $matched.Groups[1]
+    $newSHA256 = $matched[0].Groups[1]
     if ($newSHA256 -eq "") {
         Write-Warning "Could not find $image in checksums file"
         return $false
--- a/.github/actions/download-rootfs/download-rootfs.ps1
+++ b/.github/actions/download-rootfs/download-rootfs.ps1
@@ -63,9 +63,14 @@ function Test-Checksums {
 
     # Parse checksum file
     $image = $URL.Segments[$Url.Segments.Length - 1]
-    $pattern = "(\w+)  ${image}"
+    $pattern = "(\w+) \*${image}"
+    $matched = (Select-String -Pattern "${pattern}" -Path "${Checksums}").Matches
+    if ($null -eq $matched -or $matched.Count -eq 0) {
+        Write-Warning "Could not find $image in checksums file"
+        return $false
+    }
 
-    $newSHA256 = (Select-String -Pattern "${pattern}" -Path "${Checksums}").Matches[0].Groups[1]
+    $newSHA256 = $matched.Groups[1]
     if ($newSHA256 -eq "") {
         Write-Warning "Could not find $image in checksums file"
         return $false
--- a/e2e/launchertester/basic_setup_test.go
+++ b/e2e/launchertester/basic_setup_test.go
@@ -82,6 +82,7 @@ func TestSetupWithCloudInit(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			if tc.withWSL1 {
 				require.NoError(t, exec.Command("wsl.exe", "--set-default-version", "1").Run(), "Setup: Cannot set WSL1 as default version")
+				require.NoError(t, exec.Command("wsl.exe", "--shutdown").Run(), "Setup: Cannot enforce WSL1 as default version y shutting down the VM")
 				t.Cleanup(func() {
 					t.Log("Cleaning up: Setting WSL2 back as default version")
 					require.NoError(t, exec.Command("wsl.exe", "--set-default-version", "2").Run(), "Setup: Cannot set WSL2 back as default version")
--- a/e2e/launchertester/basic_setup_test.go
+++ b/e2e/launchertester/basic_setup_test.go
@@ -58,7 +58,8 @@ func TestSetupWithCloudInit(t *testing.T) {
 		"With only remote users":  {wantUser: "testmail"},
 		"With broken passwd file": {wantUser: "testmail"},
 		"Without checking user":   {install_root: true, wantUser: "root", wantFile: "/home/testuser/with_default_user.done"},
-		"Do not block on WSL1":    {install_root: true, withWSL1: true, wantUser: "root"},
+		// TODO: Investigate why this causes the CI VM to crash and reenable it.
+		//"Do not block on WSL1":    {install_root: true, withWSL1: true, wantUser: "root"},
 	}
 
 	home, err := os.UserHomeDir()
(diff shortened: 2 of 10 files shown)


.github/actions/download-rootfs/download-rootfs.ps1
```diff
@@ -63,9 +63,14 @@ function Test-Checksums {
...
-    $pattern = "(\w+)  ${image}"
+    $pattern = "(\w+) \*${image}"
+    $matched = (Select-String -Pattern "${pattern}" -Path "${Checksums}").Matches
+    if ($null -eq $matched -or $matched.Count -eq 0) {
+        Write-Warning "Could not find $image in checksums file"
+        return $false
+    }
 
-    $newSHA256 = (Select-String -Pattern "${pattern}" -Path "${Checksums}").Matches[0].Groups[1]
+    $newSHA256 = $matched.Groups[1]
```
ashuntu, line 73: I'm not very familiar with powershell, is there a reason `$matched` effectively went from `...Matches[0]` to just `...Matches`?

e2e/launchertester/basic_setup_test.go
```diff
@@ -45,6 +46,7 @@ func TestSetupWithCloudInit(t *testing.T) {
 	testCases := map[string]struct {
 		install_root     bool
 		withRegistryUser string
+		withWSL1         bool
```
ashuntu, line 49: Am I missing something or is this never set to true in the below test cases?

.github/actions/download-rootfs/download-rootfs.ps1
CarlosNihelton: Me messing with the working code during rebases...

e2e/launchertester/basic_setup_test.go
CarlosNihelton: I think I lost something during rebases :)

ashuntu (approved):
This looks good to me now :)

CarlosNihelton:
That was weird:

> The self-hosted runner: wsl-ci lost communication with the server. Verify the machine is running and has a healthy network connection. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error. [Source](https://github.com/ubuntu/WSL/actions/runs/12379810772/job/34554891321?pr=508)

CarlosNihelton:
The VM was crashing. There are quite a few events like the one below that happened yesterday and today.

image

I resized it to the [second VM size compatible with GitHub large Windows runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#specifications-for-general-larger-runners) (which are available by default now for open source projects) so that, if we were to migrate this out of our self-hosted runners we wouldn't be surprised with a downgrade. I then resized to the first size, as the change didn't solve the crash. That seems outside of our control. Even with more powerfull specs than my laptop, the crash still happened in the VM but didn't in my computer.

I'll turn the WSL1 test case off for now and park a card to investigate this later.

ashuntu (approved):
Interesting issue with the VMs, something to investigate later. Everything else looks good

PR #513, opened 2025-02-12
Migrate to org GitHub to Jira application

didrocks:
We have a new GitHub to Jira global organization application. The previous gh workflow action is now deprecated.

UDENG-5945


CarlosNihelton (approved):
LGTM!

End-to-end tests won't pass because we no longer have current images in c-i.u.c.
Let's park that work for a subsequent PR.

PR #514, opened 2025-02-21
feat: Support modern WSL installs in CI

ashuntu:


CarlosNihelton (approved):
It doesn't seem to break GoWSL, but will likely break UP4W. Not an issue, though. We can assess and fix UP4W subsequently.

PR #517, opened 2025-11-20
feat(wsl-checkout): Add support for checking out a private repository into WSL

hk21702:
This PR adds support for checkout out a private repository into WSL by utilizing a GitHub token.

Note that the token persists in the git config even after the action completes. Since we aren't using a JavaScript action, we don't have the ability to cleanup the token in the git configuration at the end. Additionally, clearing the git configuration at the end of the main action would cause a change in behavior down the road between private and public repositories at best, and a breaking behavior change at worst. Thus, we only just warn the user in the description, and by default not use any token at all.

As a consequence, one should only be using this feature in tightly controlled environments such as self-hosted runners and should also use scoped permissions. Presently, the existing e2e tests only handle the non-token case for this reason.

Realistically, we should probably overhaul many of these actions to be more robust at some point.

### DIFF
--- a/.github/actions/wsl-checkout/action.yaml
+++ b/.github/actions/wsl-checkout/action.yaml
@@ -1,5 +1,5 @@
 name: Azure WSL repo clone
-description: 'Clones your repo into WSL.'
+description: "Clones your repo into WSL."
 
 inputs:
   distro:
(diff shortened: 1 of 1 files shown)


edibotopic:
Thanks @hk21702 --- it's also a good opportunity to revisit that documentation in general.

cnihelton (approved):
Good job! We can ignore the appx build errors for now, as they have nothing to do with this PR.

PR #518, opened 2025-12-03
maint(ci): Disable scheduled workflow runs

cnihelton:
As we're focused on the tar-based format, there are issues with the scheduled jobs we cannot address now.
Let's disable them for good.


jibel (approved):
:+1:

PR #522, opened 2026-06-18
feat: Update Windows Terminal profile fragment

cnihelton:
To get in sync with https://github.com/ubuntu/wsl-setup/pull/49, including a light theme.
We still cannot ship fonts in packaged apps, though.
Also in recent versions of the Windows Terminal the asset icon is not being loaded from the web.
But the issue mentioned in the comments is fixed since late 2025, so we can confidently point to the icon present in the package itself.

---

UDENG-10627


hk21702 (approved):
LGTM

PR #531, opened 2026-08-18
Fix ARM64 intrinsics error due outdated toolset

cnihelton:
The Windows SDK version installed in CI calls intrinsics not supported in the v142 toolset, thus we need to upgrade:

ClCompile target) ->
         C:\Program Files (x86)\Windows Kits\10\Include\10.0.26100.0\um\winnt.h(6343,12): error C3861: '_CountOneBits64': identifier not found [C:\Temp\builddir\DistroLauncher\DistroLauncher.vcxproj]
         
Here's a failed CI build attempt: https://github.com/ubuntu/WSL/actions/runs/32178337610

Alternatively we could change the GH runner to `windows-2022`and hardcode one of the older SDK versions for the time they remain available: 

> Installed Windows SDKs
> - 10.0.17763.0
> - 10.0.19041.0
> - 10.0.22621.0

Keeping the project buildable on the latest seems more sustainable.

### DIFF
--- a/DistroLauncher-Appx/DistroLauncher-Appx.vcxproj
+++ b/DistroLauncher-Appx/DistroLauncher-Appx.vcxproj
@@ -35,26 +35,26 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v142</PlatformToolset>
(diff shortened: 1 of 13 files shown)


cnihelton:
Tested the sideload version from the build https://github.com/ubuntu/WSL/actions/runs/32261101807 on a Windows 10 VM running the build number 10.0.19045.6456. All looking good.

hk21702 (approved):
LGTM. I wouldn't say no to pinning the Windows runner version we use to prevent this, but it also acts as a good warning that it's time to bump things, so it's probably also fine to keep it as is.

